PR #4, opened 2019-01-27
CSS overrides, added GLOSSARY.md, page navigation names

OrangeChannel:
Condensed a lot of the styles/website.css and added a GLOSSARY.md file which will be used a lot later to avoid redundancy. It's documented working with this version of gitbook. There is a small bug with shadows on theme-1 (sepia) clipping but I'm not sure what's causing it. I'll add it to my to-do list.

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -1,0 +1,10 @@
+## TPP
+Timing Post-Processor, a built-in feature of Aegisub that can help automate
+some timing work, but doesn't always yield desired results.
+
+## kfx
+Karaoke effects, often in softsubs represented with \{\\k\} tags.
+
+## Dependency Control
+An automation script package manager built by fansubbers. The open-source
+repository can be found at https://github.com/TypesettingTools/DependencyControl.
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-- [Contributing](CONTRIBUTING.md)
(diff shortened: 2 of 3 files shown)


OrangeChannel:
I just checked on git and the css file has some lines indented at weird levels (not that this affects anything besides cleanliness) that are showing as correctly lined up on my editor (atom). Not sure what setting is causing that.

FichteFoll:
Thanks for the changes. What follows is a workflow suggestion for this and potential other projects.

---

This PR tackles three different topics (or has three concerns) that are not related:

1. adjusting the styling via website.css
2. adding a glossary
3. changing some page names in the sidebar

Each of these changes should get its own commit because they are topically unrelated and can stand on their own. Currently this is all happening within a single commit. You can easily see this by the fact that the commit topic (the first line of the commit message) has three items, as you deemed them all important enough to be mentioned (and with good reason).

The next step would then be to work on the different areas in separate branches and submit a PR for each of these. In case you find changes you want to amend to a branch, like the second commit here adjusting the website styling, you would add that commit to the branch where you added the other styling commit and I could review those together.

This is just a website providing some guidance for fansubbing, so it's not a huge issue of course, but I'm saying this in case you are going to do contributions to other projects and probably everyone prefers looking at and approving changes in an isolated fashion rather than having to sort things out from a single commit in their head. Think of it as training.

In case you realized this already and are looking for how to split multiple different changes you made onto the worktree into separate commits (& branches), I can give you some guidance on that as well, if you need it.

Issue #2, opened 2018-04-14
Resources


FichteFoll:
Just a collection of links that can be referenced when working on topics.
Some of these are referenced in the guide already, but I'll include them for completeness sake.

## Various

Collections of various stuff related to fansubbing, either as a collection of links or in the form of (a) guide(s).

- https://yukisubs.wordpress.com/guides/ (markdown: https://gist.github.com/YukinoAi/acea024631a2585aa592b16b4bde959f)
- https://subarashii-no-fansub.github.io/Subbing-Tutorial/ (blender, ocr, xdcc (lol?), xdelta)
- https://tilde.club/~garret/fansub.html

## Editing

- https://collectr.blogspot.com/2013/10/editorial-minimalism.html
- https://collectr.blogspot.com/2011/01/collectrs-curmudgeonly-guide-to-editing.html
- http://m33w-fansubs.com/editing-guide

## Timing

- https://scribbles.moe/timing-with-skiddiks/ ([archive.org](https://web.archive.org/web/20150526010125/http://scribbles.moe/timing-with-skiddiks/))
- https://commiesubs.com/wp-content/uploads/2011/06/A-guide-to-timing-in-Aegisubv2.pdf
- http://doki.co/support/doki-timing-guide/

## Typesetting

- http://unanimated.hostfree.pw/ts/?i=1
- ~~http://blog.line0.in/typesetting-with-illustrator-and-ai2ass-part-i-the-basics/~~ (migrated to the site below)
- https://typesettingtools.github.io/
- gradient along clip edge (by GrygrFlzr) https://cdn.discordapp.com/attachments/500474187568250912/505096049862508547/unknown.png (source pdf missing)

## Encoding

- https://www.animemusicvideos.org/guides/avtech31/
- https://encoding-guide.neocities.org/
- http://vsdb.top/
- https://github.com/vapoursynth/vsrepo
- deblocking: https://people.xiph.org/~xiphmont/demo/daala/demo1.shtml
- introduction to color spaces: https://ciechanow.ski/color-spaces/
- informative articles on resampling: https://entropymine.com/imageworsener/
- visual explanation of common dithering algos: http://caca.zoy.org/wiki/libcaca/study/1 + http://caca.zoy.org/wiki/libcaca/study/2 + http://caca.zoy.org/wiki/libcaca/study/3 + http://caca.zoy.org/wiki/libcaca/study/4
[...]

PR #5, opened 2019-01-30
Issue #2 addressed, semantic linefeeds, added guidelines for hyperlinking and sourcing

OrangeChannel:
Addressed #2 and added sections for hyperlinking and archiving websites that are probably not permanent.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,39 +54,26 @@ why it only applies to a specific language.
 This guide is written in [Markdown][]
 and uses [gitbook][]'s [toolchain][] to compile
 the static HTML pages.
-gitbook uses the [Github Flavoured Markdown][GFM] (GFM) variant.
+Gitbook uses the [Github Flavoured Markdown][GFM] (GFM) variant.
 
-Documentation for gitbook's toolchain can be found here:
+Documentation for gitbook's toolchain (including gitbook markdown sytax)
+can be found here:
 <https://toolchain.gitbook.com/>
 
 In order to build and preview the guide locally,
 you need [npm][] and [node.js][].
 The former is usually bundled with installation packages for node.
-Once you got those installed,
+Once you have those installed,
 run the following commands
 from the repository's folder:
 
-{% term %}
+```js
 $ npm install
-added 611 packages from 674 contributors in 4.478s
 
 $ ./node_modules/.bin/gitbook install
-info: installing 5 plugins using npm@3.9.2
-…
 
 $ ./node_modules/.bin/gitbook serve --open
-Live reload server started on port: 35729
-Press CTRL+C to quit ...
-info: 12 plugins are installed
-info: loading plugin "highlight"... OK
-…
-info: found 11 pages
-info: found 21 asset files
-…
-info: >> generation finished with success in 1.6s !
-Starting server ...
-Serving book on http://localhost:4000
-{% endterm %}
+```
 
 Afterwards, your browser will have opened
 with a preview of your files.
@@ -114,20 +101,20 @@ for each of the roles
 specified in the roles page.
 
 Feel free to work on any of the `TODO` marks
-or on a new section.
+or create a new section.
 
 Currently, we aim to add the following topics
 in no particular priority:
 
 - Workflow
-- (Translation)
+- Translation
 - Edit
 - Timing
   - Basic Procedure
   - Snapping
   - Joining, Splitting
   - Post-processing (TPP & Useful Scripts)
-  - Shifting & Sushi
+  - Shifting & [Sushi][]
   - Karaoke
 - Typesetting
   - …with Aegisub
@@ -135,37 +122,39 @@ in no particular priority:
     - Signs
       - Positioning, Layers, Rotation, Perspective, …
     - Masking
-    - Automation Scripts
+    - [Automation Scripts][TypesettingTools]
     - Movement & Motion Tracking
-  - …with Adobe Illustrator
+  - …[with Adobe Illustrator][Ai2ASS]
   - (…with Adobe After Effects)
 - Encoding \[*I'm sure there's something to be done*\]
 - Quality Check
 - Karaoke Effects
 
-You may find some inspiration for these topics
-from the following resources:
+There is a collection of links [here][issue2] that can be used as reference
+when working on any future section.
 
-- http://unanimated.hostfree.pw/ts/?i=1
-- http://blog.line0.in/typesetting-with-illustrator-and-ai2ass-part-i-the-basics/
-- https://commiesubs.com/wp-content/uploads/2011/06/A-guide-to-timing-in-Aegisubv2.pdf
-- http://doki.co/support/doki-timing-guide/
+[Ai2ASS]: https://typesettingtools.github.io/2014/08/25/typesetting-with-illustrator-and-ai2ass-part-1.html
+[TypesettingTools]: https://github.com/TypesettingTools
+[Sushi]: https://github.com/tp7/Sushi/releases
+[issue2]: https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/issues/2
 
 
 ## Style Guidelines
 
-Following are the style guidelines
+The following are the style guidelines
 for various aspects of this guide.
-The most important aspect are **Semantic Linefeeds**.
-All the other points
-serve as guidelines to refer to
-in case you are wondering
-how to format Stuff™.
+The most important aspect is having **Semantic Linefeeds**.
+The other points may serve as guidelines for formatting future pages.
+Refer to [gitbook's markdown documentation][gitbookMarkdown]
+for guidelines on visual formatting.
+
+[gitbookMarkdown]: https://toolchain.gitbook.com/syntax/markdown.html
 
 
 ### Semantic Linefeeds (!)
 
-Always use [Semantic Linefeeds][] are used
+Always use [Semantic Linefeeds][] when editing code.
+They are used
 to break lines into logical units
 rather than after a certain line length threshold is reached!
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -114,58 +101,60 @@ for each of the roles
...
+Refer to [gitbook's markdown documentation][gitbookMarkdown]
+for guidelines on visual formatting.
+
+[gitbookMarkdown]: https://toolchain.gitbook.com/syntax/markdown.html
 
 
 ### Semantic Linefeeds (!)
 
-Always use [Semantic Linefeeds][] are used
+Always use [Semantic Linefeeds][] when editing code.
```
FichteFoll: It's text, not code.

CONTRIBUTING.md
```diff
@@ -186,7 +175,7 @@ a comma,
 a period,
 any other sentence terminating punctuation,
 parenthesized sentences (not words),
-long lists (like this).
+or new items in a long list (such as the one you are reading right now).
```
FichteFoll: nitpick: I'd put the parenthesized block on its own line

CONTRIBUTING.md
```diff
@@ -198,21 +187,22 @@ The indent size is **two spaces**.
 
 ### Lists
 
-Unordered lists should be indented **once**,
+Unordered list lines should be indented **once**,
 while ordered lists are indented **twice**.
-This also applies to the first line of an ordered list,
-so that there are two spaces after the first item.
+The text of an unordered item should have one space after the `-`
+while the text of an ordered item should have two spaces after the `#.`
```
FichteFoll: This isn't true in general. When a list reaches 10 items (and you write the numbers out instead of just repeating `1.`), it would have a single space after `##.`
I guess you could say that the text of numbered list items begins after column 4.

CONTRIBUTING.md
```diff
@@ -198,21 +187,22 @@ The indent size is **two spaces**.
...
 
-
-1.  Text be here.
-    And consecutively indented with four spaces.
+```md
+1.  This is an ordered list.
+    Consecutive lines are indented with four spaces.
 2.  Another list item
-…
+...
```
FichteFoll: :thinking:

CONTRIBUTING.md
```diff
@@ -224,6 +214,66 @@ should be separated from text
...
+For relative links (links to other pages, images, or files within
+this repository), follow the guidelines for [Jekyll Relative Links][].
+
+[short]: https://guide.encode.moe/
+[Jekyll Relative Links]: https://github.com/benbalter/jekyll-relative-links/blob/master/README.md
+
+
+### Citations
+
+If you are archiving another website's text or copying their images
```
FichteFoll: This section isn't using semantic linefeeds.

CONTRIBUTING.md
```diff
@@ -224,6 +214,66 @@ should be separated from text
...
+[short2]: https://guide.encode.moe/CONTRIBUTING.HTML
+...
+```
+
+If you are linking to a section on the same page,
+`[section name](#header)` is allowed in-line.
+An example of this is ["the hyperlink section you are reading"](#Hyperlinking).
+In code, this is simply `["the hyperlink section you are reading"](#Hyperlinking)`.
+
+For relative links (links to other pages, images, or files within
```
FichteFoll: semantic linefeeds~

CONTRIBUTING.md
```diff
@@ -224,6 +214,66 @@ should be separated from text
...
+```md
+[short1]: https://guide.encode.moe/
+[short2]: https://guide.encode.moe/CONTRIBUTING.HTML
+...
+```
+
+If you are linking to a section on the same page,
+`[section name](#header)` is allowed in-line.
+An example of this is ["the hyperlink section you are reading"](#Hyperlinking).
+In code, this is simply `["the hyperlink section you are reading"](#Hyperlinking)`.
```
FichteFoll: Markdown isn't "code". (also why quote this?)

overview/preface.md
```diff
@@ -28,5 +28,5 @@ available to fansubbers, so if they are addressed in this guide, it will
 not be in as much depth as other roles.
 
 If you don't even know what roles there are to choose from, don't
-worry — we'll get there. For now, let's move on to some practical
+worry—we'll get there. For now, let's move on to some practical
```
FichteFoll: ~~Use an em dash (`—`) when stripping whitespace, otherwise an en dash (`–`) is acceptable.~~
Iirc, the former is the ususal English style, while the latter is used in German (and I find it visually more appealing) and also accepted in English, so I chose that. I don't have good sources on English punctuation rules, however.

Nvm, this is an em dash already.

overview/requirements.md
```diff
@@ -77,15 +77,15 @@ needs to be.
 
 Every role will have different required software, but it is recommended
 for every role to have installed [Aegisub][]. It
-is **highly** recommended to use [line0's builds][]. They include a
-preinstalled automation script package manager and several *critical*
-fixes to Aegisub that have not been merged into the official
-application.
+is **highly** recommended to use [line0's builds][].
+They come pre-equipped with Dependency Control
```
FichteFoll, line 84: Is "Dependency Control" defined/explained anywhere?

FichteFoll (changes requested):

CONTRIBUTING.md
OrangeChannel: Oh that looked like an underscore, reverted.

overview/requirements.md
OrangeChannel: Yes, it's in GLOSSARY.md . If this version of Gitbook is documented correctly, it should become a highlighted phrase that when you mouse over explains what it is with a link to the section in the glossary. This was mainly a test of the feature, but I could see it being used in the future to avoid having footnotes and redundancy when explaining something off-topic.

CONTRIBUTING.md
OrangeChannel: Although, I'm confused, why does Atom render both as the same size and spacing? https://imgur.com/VkrS38q and refer to https://en.wikipedia.org/wiki/Ellipsis / https://blog.apastyle.org/apastyle/2015/05/punctuation-junction-quotation-marks-and-ellipses.html . I have always written papers in APA format so it's just a habit to not use the unicode character here.

CONTRIBUTING.md
OrangeChannel: Good way of wording it.

CONTRIBUTING.md
```diff
@@ -190,8 +190,9 @@ The indent size is **two spaces**.
 
 Unordered list lines should be indented **once**,
 while ordered lists are indented **twice**.
-The text of an unordered item should have one space after the `-`
-while the text of an ordered item should have two spaces after the `#.`
+The text of an unordered item should have one space after the `-`,
+while the text of an ordered item
+should start four columns after the number.
```
FichteFoll, line 195: That's still kind of ambiguous for multi-digit numbers, but we specifically have an example for that below, so ┐(°ヮ°)┌.

FichteFoll (approved):

OrangeChannel:
Adjusted the other docs to fit semantic linefeed standards so Git will stop showing giant red and green blocks, no way to do it the first time though without these blocks.

Just read your comment on the other PR (thanks for the help by the way 👍 ) and I'm wondering: Is this PR fine by those standards then?

 - I split the changes categorically this time into two commits and the PR as whole only deals with the docs under the overview SUMMARY.md header.

FichteFoll:
>Is this PR fine by those standards then?

No, because in the first commit you also changed the [install instructions](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/5/commits/c5c7e8e09bac489d1ce5605eb36c6f357bf336d4#diff-6a3371457528722a734f3c51d9238c13R70)

Otherwise, yes, this was mostly a stylistic change (like adding semantic linefeeds, a bit of rewording of some section) and adding something to the style guidelines. The commits are generally also well-separated and a commit to "address feedback" from the PR review is fine with me, too.

OrangeChannel:
Oh yea, that was a weird one, because I thought it was using the term plugin incorrectly. It's meant for better color display (which is not a bad thing) and also the ability to copy the commands line by line. The author of the original left in the \$ which should actually be added through CSS via [counter-increment](https://www.w3schools.com/cssref/pr_gen_counter-increment.asp) if you want it, because this will not be copied. You can change it back to the term format by replace the ` ```js ` with ` ```term `, but I'm not sure if those colors are supported on all three themes, or if it overwrites the background.

Also, I think (could be wrong) you have to hide the other copy button if you want to use term's, I'm not sure how to go about that besides in-line.

OrangeChannel:
https://github.com/CLIUtils/gitbook-plugin-term#block-based-auto-colorization maybe I'm interpreting this wrong though, let me know.

FichteFoll:
Iirc the `$` was required so the plugin would recognize the line with the command, as I don't exactly see how it would otherwise. And the docs seem to support that, too, and the command was picked up by the regex as well. Did you actually try it out before you removed it?

OrangeChannel:
I did but now thinking of it, I definitely didn't do it correctly. I hadn't actually set up the book but I was using a plug-in on my editor. I see it now working.

I'll add it back to how it was before, sorry about this one 😞 . Thought I was fixing it.

FichteFoll:
I still have a couple tabs open with potentially useful plugins from back then when I set up this repo. Somehow people made a lot of plugins for gitbook.

I should finally go through them at some point.

OrangeChannel:
Hey, turns out my plug-in wasn't 100% functioning wrong. <https://guide.encode.moe/CONTRIBUTING.html> checkout the first code box. The copy line *is* only showing up once but at least it is replacing the default one this time. On firefox, the copy button copies all three commands 🤷‍♀️ .

Issue #1, opened 2018-03-24
Issue about x264 in the guide


eXmendiC:
![](https://cdn.discordapp.com/attachments/224104675031384065/427062348792987649/unknown.png)
```
[12:13] Yumifumie: 8bit and 10bit are together now (so there aren't two versions anymore) and mods are too old
```

kmod:
```
[12:42] Yumifumie: 2 versions old and probably won't get an update (newest version broke some mods)
```
![](https://cdn.discordapp.com/attachments/224104675031384065/427069639348977664/unknown.png)

tmod:
```
[12:42] Yumifumie: !!!!!!WARNING!!!!!!!
Finaly the day came. The 8/10 bits unification commit breaks too much things, and some other commits, use of new ffmpeg API breaks compatibility with some patches, for which i'm obliged to stay with ffmpeg 2.8.x. So, now, even if the 7z files and the version is tagged with the actual master release version, it's not realy it.
[12:43] Yumifumie: >for which i'm obliged to stay with ffmpeg 2.8.x
>ffmpeg 2.8.x
```
![](https://cdn.discordapp.com/attachments/224104675031384065/427070027909431312/unknown.png)

PR #8, opened 2019-02-05
Encoding folder update

OrangeChannel:
Mostly just semantic linefeed updates, for future cleaning and adding. Also added [blocky artifact pitcure and additional explanation](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/commit/1b289548c6b78fe69e3114845f5404b073b63f42#diff-05e4281dac32bbfc72f55d339784eda4R61). 

- a4c6231 added Discord link for questions
- 1ac7c94 mentions #1 but I'm unsure actually what needs resolving
- 523a2ec adds some links to relevant codecs
- f33c2be references an interesting issue on waifu2x upscaler that may get confused for blocking artifacts


encoding/video-artifacts.md
```diff
@@ -58,8 +58,25 @@ disaster.
 DCT block based video compression has come a long way. If you happen
 to be encoding an MPEG2 TV station or a DVD from a previous decade, you
 will likely come across something like this:
+![Blocky Compression](images/blocky2.png)
```
FichteFoll, line 72: I believe we want to separate images (that are used like figures) with an empty from paragraphs.

encoding/video-artifacts.md
```diff
@@ -122,6 +139,7 @@ interlacing-related aliasing it is rarely recommendable.
 Ringing is something of a blanket term for edge artifacts, including
 mosquito noise, edge enhancement artifacts, overshoot, or actual
 ring-like ringing caused by the Gibbs phenomenon.
+![Mosquito Noise](images/mosquito1.png)\(Urban, 2017\)
```
FichteFoll, line 183: same

encoding/preparation.md
```diff
@@ -286,3 +286,5 @@ Just extract it somewhere.
 ---
 
 [^1]: It should be noted that the author strongly disagrees with this sentiment. The two have a lot in common, and any capable Avisynth encoder could reach a similar leven in Vapoursynth within a few months, maybe even weeks. At least I'm honest, okay?
+
+[^2]: https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/issues/1
```
FichteFoll, line 299: Instead of adding a link to this, we should just fix the text.

FichteFoll (commented):
Thanks for the changes. 

There are quite a few issues about the content that even I as a non-encoder can spot to be wrong, so it will need a pass or two, but every change is an improvement at this point.

Left a few comments for consideration before I merge.

FichteFoll:
Also, good job with the separation of concern with these PRs :+1:

encoding/video-artifacts.md
OrangeChannel: It is actually separated, it's just the way that commit chose to show the change. The space above it was considered part of the linefeeds commit. You can see the final file [here](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/8/files#diff-05e4281dac32bbfc72f55d339784eda4R67).

encoding/video-artifacts.md
OrangeChannel: Refer to [preivous answer](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/8#discussion_r254493425).

encoding/preparation.md
OrangeChannel: Agreed, but not really sure what the author of https://github.com/jpsdr/x264/releases meant. I didn't want to put something wrong there because I'm not sure exactly what happened on the release.

FichteFoll: Well, I guess something is better than nothing.

PR #10, opened 2019-02-21
Yet another encoding folder update.

OrangeChannel:
Notable changes:
- 7d1daf7 resolves #1 by updating the information regarding x264 codec (thanks kageru!)

Fixes:
- b8c98b3 sadly footnotes refuse to work with semantic linefeeds
- facbf0b + c4587c1  ffmpeg and SVT link now work
- 96566fb + c4e83e4 prevent gitbook from attempting to download a blank .md file

Formatting:
- 2e5e5a2 improved alignment for when images have a (reference)

PR #15, opened 2019-02-25
Added information about audio 

rokweom:


encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
 
 ## Audio Codecs
 
-TODO: FLAC, AAC, maybe mention opus
+### Audio formats and how to handle them
+
+Depending on the source you'll be working with,
+you may encounter many different audio formats.
+
+On Blu-Rays,
```
FichteFoll:
```suggestion
On Blu-rays, 
```

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+
+Depending on the source you'll be working with,
+you may encounter many different audio formats.
+
+On Blu-Rays, 
+you will most likely find audio encoded losslessly, 
+in the form of either DTS-HD Master Audio, Dolby TrueHD, or PCM. 
+DTS-HD MA and Dolby THD are proprietary codecs that use lossless compression, 
+while PCM is simply raw, uncompressed PCM data.
+The usual way handle these formats is to reencode them to other formats -
```
FichteFoll:
```suggestion
The usual way to handle these formats is to reencode them to other formats - 
```

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+Depending on the source you'll be working with,
+you may encounter many different audio formats.
+
+On Blu-Rays, 
+you will most likely find audio encoded losslessly, 
+in the form of either DTS-HD Master Audio, Dolby TrueHD, or PCM. 
+DTS-HD MA and Dolby THD are proprietary codecs that use lossless compression, 
+while PCM is simply raw, uncompressed PCM data.
+The usual way handle these formats is to reencode them to other formats - 
+either lossless or lossy, depending on your taste.
```
FichteFoll: This is an example where the trailing whitespace would create a line break. Not sure if intended.

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+
+On Blu-Rays, 
+you will most likely find audio encoded losslessly, 
+in the form of either DTS-HD Master Audio, Dolby TrueHD, or PCM. 
+DTS-HD MA and Dolby THD are proprietary codecs that use lossless compression, 
+while PCM is simply raw, uncompressed PCM data.
+The usual way handle these formats is to reencode them to other formats - 
+either lossless or lossy, depending on your taste.   
+But first, you need to decode them. 
+The recommended tool for that is ffmpeg,
```
FichteFoll: Why not link ffmpeg here as well? Or add it to the glossary.

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+### Which codecs to use?
+
+Once you have your lossless files decoded, you need to encode them. 
+Depending on your taste, you may choose a lossy or lossless codec. 
+The two most widely accepted codecs in fansubbing community are FLAC (lossless)
+and AAC (lossy), but recently opus (also lossy) is gaining some popularity, too.
+
+The recommended encoder for FLAC is the official one, 
+and this is the one this guide fill focus on. 
+You can download it [here][FLAC].
```
FichteFoll: Please avoid using `here` as the hyperlink text. See my explanation in [this other PR](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/10#discussion_r259413046) (I should probably add it to `CONTRIBUTING.md`).

I won't repeat this for the other occurances.

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+### Lossless or lossy?
+
+This is entirely dependent on you. 
+Some people like the idea of having an (theoretically) perfect copy of the master audio file, 
+don't mind the increase in size, 
+and state that lossless is the only way to go when archiving.
+Others prefer smaller file sizes, 
+knowing that the difference —
+assuming high enough bitrate —
+won't be audible anyway.
```
FichteFoll: Another line break occurance.

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+Formats like Monkey's Audio and TAK provide very little gain over FLAC,
+while not being as widely supported,
+and — in the case of TAK —
+closed source. 
+DTS-HD MA and Dolby THD are much less efficient than FLAC,
+and are also closed source.
+MP3 is simply obsolete,
+and Vorbis has been superseded by opus. 
+DTS and AC-3 provide even worse compression than MP3,
+and don't have any reasonable, free encoders.
```
FichteFoll, line 321: Another line break occurance.

encoding/preparation.md
```diff
@@ -250,7 +250,83 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+The two most widely accepted codecs in fansubbing community are FLAC (lossless)
+and AAC (lossy), but recently opus (also lossy) is gaining some popularity, too.
+
+The recommended encoder for FLAC is the official one, 
+and this is the one this guide fill focus on. 
+You can download it [here][FLAC].
+
+The recommended and most widely used AAC encoder is qaac, 
+which you can download [here][qaac], 
+and this is the one this guide will focus on.
```
FichteFoll: Another line break occurance.

FichteFoll (changes requested):
Thank you for your contribution. 

I didn't notice any factual errors, but I'm not an export on the topic, so I'm entrusting this to peer review. :wink: 

Many lines have trailing whitespace. I suggest to tweak your editor to trim these on save, even for Markdown, because even in the event you'd want a line break (which requires two trailing spaces), you can still use `<br>` explicitly. Line breaks should be used sparingly and in most situations you're either looking at a paragraph or some list instead.

There were a couple small errors here and there that I didn't bother pointing out as they are much easier corrected in a single pass by either myself or someone with a better grasp on the English language.

encoding/preparation.md
OrangeChannel: Is the issue here that it's a keyboard dash/hyphen and not an em-dash? I was under the impression it ignores trailing whitespace at the end of lines. 

And for this example, this would still be a problem if it does. It would make the word after the hyphen not have a space separating it from the hyphen. unlike the word "formats" and the hyphen. I'd recommend using an em dash `—` with no whitespace (can't be at the end of a line) on either side.

This would result in `...is to reencode them` \N `to other formats—either lossless or lossy, depending on your taste.`

encoding/preparation.md
OrangeChannel: There's a lot of words I should add to the glossary when I get time. For now, having it explained in-line or having it a link to the download page is fine, as I'll batch add a bunch of glossary entries at some point later.

encoding/preparation.md
OrangeChannel: For this example, use commas instead of the em-dash, as the line breaks will make this look ugly.

encoding/preparation.md
```diff
@@ -250,7 +251,85 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+
+When working with WEB and TV sources,
+you will most likely have only lossy audio available.
+The most common codecs here are AC-3, E-AC-3 and AAC.
+Lossily compressed files should generally not be reencoded -
+the proper way to handle them is to remux (i.e. copy) them to the final file.
+
+[ffmpeg]: https://www.ffmpeg.org/download.html
+
+### Which codecs to use?
```
OrangeChannel, line 284: Section headers/titles not directly under another header should have two blank lines before them, separating them from the previous section.

encoding/preparation.md
```diff
@@ -250,7 +251,85 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+and Vorbis has been superseded by opus.
+DTS and AC-3 provide even worse compression than MP3,
+and don't have any reasonable, free encoders.  
+In short - don't bother, unless you really have to, for some reason.
+
+[FLAC]: https://xiph.org/flac/download.html
+[qaac]: https://sites.google.com/site/qaacpage/cabinet
+[opus]: https://opus-codec.org/downloads/
+
+### Lossless or lossy?
```
OrangeChannel, line 330: Section headers/titles not directly under another header should have two blank lines before them, separating them from the previous section.

OrangeChannel (changes requested):
Add an empty line before the headers I marked, but everything else is looking good.

encoding/preparation.md
FichteFoll: What I was referring to is the trailing three spaces in this line. You don't see them by default, but I have https://addons.mozilla.org/de/firefox/addon/render-whitespace-on-github/?src=search for that purpose since whitespace is frequently an issue when reviewing pull requests and code on GitHub..

encoding/preparation.md
```diff
@@ -248,18 +249,102 @@ such as Intel's [SVT-AV1][] will also not be included.
...
+Depending on your taste, you can choose a lossy or lossless codec.
+The two most widely accepted codecs in fansubbing community are FLAC (lossless)
+and AAC (lossy), but recently opus (also lossy) is gaining some popularity, too.
+
+The recommended encoder for FLAC is the official one.
+Download Windows builds from [it's official site][FLAC].
+Most Linux distributions should have FLAC in their package repositories.
+
+The recommended and most widely used AAC encoder is qaac,
+available on [it's official site][qaac].
```
OrangeChannel: Wrong "it's" usage here. "Its" is the possessive form.

OrangeChannel (approved):

PR #16, opened 2019-05-13
Minor rewrites

LightArrowsEXE:
To make it flow a bit better.

### DIFF
--- a/overview/roles.md
+++ b/overview/roles.md
@@ -164,8 +164,6 @@ They are responsible for ensuring the overall quality of the
 release is up to par with the group's standards.
 They are also expected to be familiar with the workflow and many
 intricacies of every other role.
-Quality Checkers are expected to be well versed in every role
-and able to critique them.
 Each group has a different approach to how the Quality Checker
 completes their work.
 For example, one group might require an organized "QC report"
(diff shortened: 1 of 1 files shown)


overview/roles.md
```diff
@@ -161,20 +161,16 @@ own standards)
 Quality Checkers (abbreviated *QC*) are often the last eyes on an
 episode before it is released.
 They are responsible for inspecting the
-overall quality of every aspect of a show including:
-timing,
-typesetting,
-encode,
-and editing.
+ensuring the overall quality of the release is up to par
+with the group's standards.
```
FichteFoll: While I don't think we necessarily need to mention each role specifically, I do think we should clarify on the "standards" part and make a reference to all of the previous tasks as a whole.

LightArrowsEXE: Edited.

FichteFoll: Okay, `intricacies` is good for this.

However, could you reformat these sentences to use semantic linefeeds? They are explained in CONTRIBUTING.md.

LightArrowsEXE: Like so?

FichteFoll: Good enough.

PR #17, opened 2019-06-28
Merge #17: Add descaling page

LightArrowsEXE:
Wrote a post centered around how and when to descale, as well as covering how to go about it.


encoding/descaling.md
```diff
@@ -0,0 +1,356 @@
...
+usually Blu-rays,
+aren't lossless.
+There's always going to be some differences
+from the original masters
+which makes it impossible
+to perfectly descale something.
+However, usually those differences
+are so small that they're negligible.
+If you run into a case where
+you can only get high relative errors,
```
cN3rd: Misses an emphasis on "high relative errors using/given by the script".
It might confuse beginners who have no prior knowledge of descaling.

encoding/descaling.md
```diff
@@ -0,0 +1,356 @@
...
+
+If you look in the directory
+that you ran the script in,
+you will find a new folder
+called "getnative".
+You can find the following graph
+in there as well:
+
+![Manaria_Friends-ep1_04_graph](images/descale_graph.png)
+
```
cN3rd, line 161: This is also the place to discuss what are considered to be "low relative errors" and "high relative errors", when to ignore them, and how to interpret them and the graph.

encoding/descaling.md
```diff
@@ -0,0 +1,356 @@
...
+You should be looking for a low relative error.
```
cN3rd: Emphasis on that you're looking for a low error **spike**.
As you can see in the graph, the closer you get to the full resolution of the source (e.g. 1080p) the smaller the "relative error" is.

encoding/descaling.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+The next issue is figuring out
+what was used to [upsample](resampling.md#upsampling) the show.
+By default,
+getnative.py checks with Mitchell-Netravali
+(bicubic b=1/3, c=1/3).
+However, it might have also been upscaled
+using other kernels,
+like Spline or Bilinear.
```
cN3rd: An appendix or a small section about popular kernels would be useful for beginners.

encoding/descaling.md
```diff
@@ -0,0 +1,356 @@
...
+import kagefunc as kgf
+import fvsfunc as fvf
+core = vs.core
+
+src = core.lsmas.LWLibavSource("BDMV/[BDMV][190302][マナリアフレンズ I]/BD/BDMV/STREAM/00007.m2ts")
+src = fvf.Depth(src, 32)
+
+Y, U, V = kgf.split(src)
+width = vsutil.get_w(878)
+height = 878
```
cN3rd, line 329: Reversing the lines shows that getnative outputs height and you work around that:

```Py
height = 878
width = vsutil.get_w(height)
```

LightArrowsEXE: Changed

encoding/descaling.md
LightArrowsEXE: Expanded greatly on how to read the graphs, as well as showing some examples.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
+# Descaling
+
+The ability to descale is a wonderful tool
+to have in any encoder's arsenal.
+You may be familiar with
```
kageru: I don’t like how this sounds. “You may have heard before that most…” or similar.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+how most anime are not native 1080p,
+but a lower resolution.
+But how do we make use of that?
+How do you find the native resolution
+and reverse the upscale?
+
+
+## When and where to descale
+
+There's a lot of opportunities
```
kageru: “There are many circumstances under which…”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+where descaling might prove beneficial.
+For example say you've got
```
kageru: comma after example

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+a very blurry Blu-ray source.
+Rather than sharpening it,
+you might want to consider
+checking if it's possible to descale it
+and maybe alleviate a lot of
+the blurriness that way.
```
kageru: isn’t blurriness just… blur?

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+Or the opposite:
+say you've got a source full of ringing.
+It might've been upscaled
```
kageru: Please use “might have”. This isn’t commie.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
+# Descaling
```
kageru, line 1: General thing: search and replace `'` with proper apostrophes (`’`).

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+so it's worth a try
+to see if it can be descaled.
+It's no surprise that descaling tends to
+offer far better lineart
+than usual resampling does.
+
+However, descaling is not always an option.
+The worse your source is,
+the less likely it is that
+descaling will be a better alternative
```
kageru: “will yield better results than a simple resample”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+to a simple resample.
+If you've got a source
+full of broken gradients
+or noise patterns,
+like your average simulcast stream,
+descaling might only hurt the overall quality.
+Sadly, sources with a lot of post-processing
+might also prove tough to properly descale
+without dabbling into specific masks.
```
kageru: “dabble” means “to work without any deep involvement” (MW), but specific masks imply deep involvement. It’s also “dabble with”.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+However, so long as you've got
```
kageru: “as long as”. So long as is [dying out](https://books.google.com/ngrams/graph?content=as+long+as%2Cso+long+as&year_start=1800&year_end=2008&corpus=6&smoothing=5&share=&direct_url=t1%3B%2Cas%20long%20as%3B%2Cc0%3B.t1%3B%2Cso%20long%20as%3B%2Cc0#t1%3B%2Cas%20long%20as%3B%2Cc0%3B.t1%3B%2Cso%20long%20as%3B%2Cc0) and may even feel unnatural to some readers.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+a source with nice, clean lineart,
+descaling might be a viable option,
+and possibly nullify the need
+of a lot of other filtering.
```
kageru: “need for” feels more natural to me, even though [both seem to be correct](https://english.stackexchange.com/questions/6623/need-of-vs-need-for).

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+
+
+## Preparation
+
+To prepare for figuring out the native resolution,
+you'll want to use [getnative][],
```
kageru, line 58: Can we not use a hyperlink here directly?

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+a Python script designed around
```
kageru: It’s not just designed *around* that, it’s designed *for* this one purpose. It does nothing else.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+make sure to grab BluBb-mADe's [descale][].
+
+One important thing
+to keep in mind when descaling
+is that you will never find
+"the perfect descaling settings".
+This is because the best available sources to consumers,
+usually Blu-rays,
+aren't lossless.
+There's always going to be some differences
```
kageru: There *are* going to be differences (plural)

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+a Python script designed around
+figuring out the resolution
+a show was animated at.
+For the actual descaling,
+make sure to grab BluBb-mADe's [descale][].
+
+One important thing
+to keep in mind when descaling
+is that you will never find
+"the perfect descaling settings".
```
kageru, line 68: 1. Proper quotes `“”`.
2. Even though the source isn’t perfect, you can still find perfect settings (the exact same settings the studio used), so this sounds misleading to me.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+which makes it impossible
+to perfectly descale something.
+However, usually those differences
+are so small that they're negligible.
+If you run into a case where
+you can't find any low relative error spikes,
+descaling can be highly destructive.
+It's instead recommended to downscale
+as you would normally,
+or to not mess with scaling at all.
```
kageru, line 87: I’d prefer “resizing” to make it less repetitive with “descaling”.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+[getnative]: https://github.com/Infiziert90/getnative
+[descale]: https://github.com/BluBb-mADe/vapoursynth-descale
+
+
+## Finding out the native resolution
+
+To figure out what
+the native resolution of an anime is,
+first you need a good frame to test.
+Ideally you'll want a bright frame
```
kageru: Ideally,

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+with as little blur as possible of high quality
+(Blu-ray or very good webstreams).
+It also helps to not have
+too many post-processed elements in the picture.
+Whilst it is most definitely possible
+to get pretty good results with "bad" frames,
+you'll want to generally lower that chance
```
kageru: Lower what chance? The sentence isn’t really coherent.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+as much as possible.
+
+Here's some examples of "bad" frames.
```
kageru: Here are

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+
+![Manaria Friends Ep01 frame 1](images/descale_manaria01.png)
+
+This picture is dark.
+It also has some effects over it.
+
+![Manaria Friends Ep01 frame 2](images/descale_manaria02.png)
+
+This picture is also very dark,
```
kageru: no comma here

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+
+```
+Using imwri as source filter
+501/501
+Kernel: bicubic AR: 1.78 B: 0.33 C: 0.33
+Native resolution(s) (best guess): 878p
+done in 18.39s
+```
+
+If you look in the directory
```
kageru: maybe “check“ instead of “look in”?

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+that you ran the script in,
```
kageru: “where you executed the script”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+you will find a new folder
+called "getnative".
+You can find the following graph
+in there as well:
+
+![Manaria_Friends-ep1_04_graph](images/descale_graph.png)
+
+
+The X axis shows the resolutions that were checked
```
kageru: comma after checked

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+and the Y axis shows the relative error.
+The relative error refers to
+the difference between the original frame
+and the rescaled frame.
+What you're looking for
+are the low spikes,
+as that shows a low relative error.
```
kageru: Any low value is a low relative error. Higher resolutions tend to have lower errors because less information is lost or changed during de- and rescaling. We need to make clear that the spikes break this pattern by having a much lower error rate compared to similar resolutions, not compared to all tested resolutions.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+In this case it very clearly points to 878p.
+
+As a sidenote,
+it's important to note that this script
```
kageru: “sidenote … to note” sounds repetitive

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+can't find native 1080p elements.
+This is because it descales the frame
+and reupscales it afterwards
+to determine the relative error.
+You can't descale to 1080p
+if the frame is already 1080p.
+If you have reason to believe
+that your show might be native 1080p,
+you've got to go with your gut.
```
kageru, line 181: Add an example of a descale graph for a native 1080p frame.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+Output:
+Kernel: bicubic AR: 1.78 B: 0.33 C: 0.33
+Native resolution(s) (best guess): 1000p, 974p, 810p
+```
+
+This graph has a lot of unnatural swerves
+and it's impossible to determine
+what the native resolution is.
+
+Another pitfall you can fall in
```
kageru: “pit*fall* you can *fall* in”? :thinking:

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+let's look at a good graph.
+
+![Aikatsu_Friend_NCOP_graph](images/descale_good_graph.png)
+```
+Output:
+Kernel: bicubic AR: 1.78 B: 0.33 C: 0.33 
+Native resolution(s) (best guess): 810p
+```
+The results are very clear.
+There's a couple of deviations,
```
kageru: There are

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+for testing out varying kernels.
+
+## Descaling
+
+Now it's time to actually start descaling.
+Open up your Vapoursynth editor of choice,
+and import the clip:
+
+```Py
+src = lvf.src(r"BDMV/[BDMV][190302][マナリアフレンズ I]/BD/BDMV/STREAM/00007.m2ts")
```
kageru: Don’t use your wrapper just to import something in these examples. A regular lsmash call will do just fine.
Also remove the `r` before the string as it’s not needed when using `/` as a path separator.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+The next issue is figuring out
+what was used to [upsample](resampling.md#upsampling) the show.
```
kageru: It seems odd to me that we’re talking about de*scaling*, re*scaling*, down*scaling* but now up*sampling*. I’d just call this link upscale, even if it links to upsampling to keep terminology consistent within this section.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+getnative.py checks with Mitchell-Netravali
+(bicubic b=1/3, c=1/3).
+However, it might have also been upscaled
+using other kernels.
+
+Here is a list
+of some common kernels and values.
+
+* Lanczos
+* Spline
```
kageru: There are multiple splines. Kekkai Sensen used Spline16 iirc, so list 16 and 36 separately.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+* Bilinear
+* Bicubic b=1, c=0 (B-Spline)
+* Bicubic b=0, c=0 (Hermite)
+* Bicubic b=1/3, c=1/3 (Mitchell-Netravali)
+* Bicubic b=0, c=0.5 (Catmull-Rom)
+* Bicubic b=0, c=1 (Sharp Bicubic)
+
+The best way to figure out
+what is used is to simply try out
+a bunch of and use your eyes.
```
kageru: There might be a word missing here. Or there’s one too many.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+Check for common scaling-related artifacting,
+like haloing,
+ringing,
+aliasing,
```
kageru: Why is there a linebreak after each word?

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+etc.
+
+For bicubic it is important
```
kageru: For bicubic,

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+to keep in mind that
+you will typically find that
+the values match the following mathmatical expressions:
+
+* `b + 2c = 1`
```
kageru, line 301: None of your examples follow the second rule, and I don’t know of any that do. I think it’s either `b + 2c = 1` or one of them is 0 and the other could be anything. We don’t need 4 rules to express that.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+* `b + c = 1`
+* `b = 0, c = X`
+* `b = 1, c = 0`
+
+Whilst this isn't a 100% guarantee,
+this is the most common approach
```
kageru: repeated “this”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+to resampling using bicubic,
+so it's worth keeping in mind.
+
+Here's an example of the previous frame
+when descaled using various kernels and settings
+(note that descale requires either GrayS, RGBS, or YUV444PS.
```
kageru, line 312: “descaled requires floating point input” what that is should be explained elsewhere, and we don’t want to update this page if another float format is added in the future.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+I'll be using `split` and `join` from `kagefunc` to split the planes
+and then join them again in this example,
+and `get_w` from `vsutil` to calculate the width)[^1]:
+
+[Comparison between frames][manaria_compare]
+
+```Py
+import vapoursynth as vs
```
kageru: You only use `core` from vapoursynth, so you can just `from vapoursynth import core`

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+import vsutil
+import kagefunc as kgf
+import fvsfunc as fvf
+core = vs.core
+
+src = core.lsmas.LWLibavSource("BDMV/[BDMV][190302][マナリアフレンズ I]/BD/BDMV/STREAM/00007.m2ts")
+src = fvf.Depth(src, 32)
+
+Y, U, V = kgf.split(src)
```
kageru: Variable names should be in lowercase

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+height = 878
+width = vsutil.get_w(height)
+
+descale_a = core.descale.Debilinear(Y, width, height).resize.Bilinear(1920, 1080)                                                        # Bilinear
```
kageru: Please put the `#comment` above the corresponding function call.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+descaled = kgf.join([descale_a, U, V])
+descaled.set_output()
+```
+
+You might realize that after descaling,
+we are immediately upscaling the frame
+with the same kernel and values again.
+This is done so we can compare the before and after.
+The closer the new frame is to the old one,
+how more likely it is that you've got the correct kernel.
```
kageru: “the more likely it is”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+Zooming in on the frame at 4x magnification or higher
+will help immensely.
+An alternative that you can do
```
kageru: “that you can use”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+descaled.set_output()
+```
+
+You might realize that after descaling,
+we are immediately upscaling the frame
+with the same kernel and values again.
+This is done so we can compare the before and after.
+The closer the new frame is to the old one,
+how more likely it is that you've got the correct kernel.
+Zooming in on the frame at 4x magnification or higher
```
kageru, line 358: Magnification using nearest neighbor. Using bilinear or bicubic on the preview won’t show you anything of value.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+src = fvf.Depth(src, 32)
+
+Y, U, V = kgf.split(src)
+height = 878
+width = vsutil.get_w(height)
+
+descale_a = core.descale.Debilinear(Y, width, height).resize.Bilinear(1920, 1080)                                                        # Bilinear
+descale_b = core.descale.Debicubic(Y, width, height, b=1/3, c=1/3).resize.Bicubic(1920, 1080, filter_param_a=1/3, filter_param_b=1/3)    # Mitchell-Netravali
+descale_c = core.descale.Debicubic(Y, width, height, b=0, c=1).resize.Bicubic(1920, 1080, filter_param_a=0, filter_param_b=1)            # Sharp Bicubic
+descale_d = core.descale.Debicubic(Y, width, height, b=1, c=0).resize.Bicubic(1920, 1080, filter_param_a=1, filter_param_b=0)            # B-Spline
```
kageru: You’re just discarding all of these clips without ever saying why. The reader can’t know that you’re changing the `kgf.join()` call between comparisons.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+descaled = kgf.inverse_scale(src, height=878, kernel='bicubic', b=0, c=1/2, mask_detail=True)
+```
+
+We can make use of the mask
+that `inverse_scale` uses internally
+as well.
+
+```Py
+descaled = kgf.inverse_scale(src, height=874, kernel='bicubic', b=0, c=1/2)
+descaled_mask = kgf._generate_descale_mask(vsutil.get_y(core.resize.Spline36(src, descaled.width, descaled.height)), vsutil.get_y(descaled), kernel='bicubic', b=0, c=1/2)
```
kageru, line 394: You’re not supposed to call methods prefixed with `_`. Use `mask_descale` instead.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+![KaguyaOP_credits_mask](images/descale_credits_mask.png)
+![KaguyaOP_descaled](images/descale_credits.png)
+
+Note that if you see the mask
+catching a lot of other stuff,
+you might want to consider *not* descaling
+that particular frame,
+or to try a different kernel/values.
```
kageru: “or trying”

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+
+There are various things you can do
+to deal with scenes that have issues
+even after descaling.
+`Eedi3` stands out in particular
+as a fantastic AA filter
+that really nails down bad lineart
+caused by bad descaling.
+Just try not to rely on it too much,
+as it is by all means a "cheat code".
```
kageru: Don’t make it sounds so good or people will think it magically fixes everything. Like descaling to the wrong resolution and then having the detail mask and eedi carry you. :smug_nep:

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+try a different kernel or values.
+Alternatively,
+try not descaling at all.
+
+At the end of the day,
+as mentioned in the introduction,
+you can't descale everything perfectly.
+Sometimes it's better to think of it
+as a magical anti-ringing/haloing/aliasing filter
+(in conjunction with a good upscaling filter
```
kageru: Make it more clear that rescaling/rpow2 magic is optional. Currently, this implies that you have to upscale again.

encoding/descaling.md
```diff
@@ -0,0 +1,452 @@
...
+scaled = nnedi3_rpow2(scaled).resize.Spline36(1920, 1080)
+```
+
+Note how this fixed
+most of the aliasing
+on the CGI model.
+
+***
+
+[^1]: Most, if not all relevant Vapoursynth scripts/plug-ins and their functions can be found in the [VapourSynth Database][vsdb].
```
kageru: “Most, if not all, relevant”

kageru (changes requested):
Note: All of my QC comments are to be understood as suggestions, not absolute, objective truth.

encoding/descaling.md
LightArrowsEXE: For this example I wanted to generate the mask specifically. As far as I can tell, I can't do that with `mask_descale`.

encoding/descaling.md
FichteFoll: What's the line where `might've` isn't accepted but `you've` is? Is there something semi-official or just your gut feeling? (Imo this is fine. Has been adressed already, though.)

encoding/descaling.md
FichteFoll: Do we want to make this a guideline? I don't particularly care about the correct use of apostrophes in a context like this.

encoding/descaling.md
FichteFoll: Hyperlinks in text are disrupting, artificially increase line length in source and work against semantic linefeeds. It's better to declare them after the paragraph (or at the end if they are used multiple times).

encoding/descaling.md
FichteFoll: Good point on 2. For 1. see my comment earlier on apostrophes.

encoding/descaling.md
FichteFoll: ^nit but valid

encoding/descaling.md
FichteFoll: i.e. reference to the docs in a footnote?

FichteFoll:
Reviewed @kageru's review and left comments where I disagreed or am indifferent. Comments I upvoted I support strongly, the others just normally. I also resolved those that have been addressed already.

I'll trust kageru's judgement for this section and wait until his review has been addressed, after which I'll do a final pass and most likely merge.

---

@cN3rd, you can combine multiple comments in a single review if you click on "Start a review" on the first comment, proceed to add more comments and finally click on the "Review Changes" button to submit your review comments together.

encoding/descaling.md
kageru: Just gut feeling. You’ve feels… acceptable(?) to me, but might’ve gives me this feeling of forced contraction just for the sake of it.

encoding/descaling.md
kageru: Nvm, I’m an idiot. I thought I put a debug parameter in there.

encoding/descaling.md
kageru: Maybe we should get a third opinion on this. I like them, but I guess most people don’t care.

encoding/descaling.md
kageru: Will this syntax insert a hyperlink on the word `getnative` using the link defined below?

encoding/descaling.md
LightArrowsEXE: Leaving a note for myself to look at this later, as I can't think of a clear way to do this yet. To me, the "spike" itself seems like an obvious enough example of the pattern being broken. Will think about it.

encoding/descaling.md
LightArrowsEXE: I was under the impression that you always had to break the line after a comma with semantic linefeeds.

LightArrowsEXE:
These two comments remain unresolved, as I don't know how to exactly go about them:

https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/17/files/2718ea784f5a17ab79dfd7e8766b5a103723c2a5#diff-90c73b01b7f851b1c04c0f924d50b478
https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/17/files/2718ea784f5a17ab79dfd7e8766b5a103723c2a5#diff-90c73b01b7f851b1c04c0f924d50b478

(also I just realized I said "discord" instead of "github" in the commit description)

encoding/descaling.md
FichteFoll: Yes, this is a very standard markdown feature and used throughout the guide already.

encoding/descaling.md
FichteFoll: The oxford comma only applies for enumerations. Is a comma actually valid here?

encoding/descaling.md
FichteFoll: You don't *have* to do this if the enumeration is short, like it is here. Dave just has a dendency to do it for all of them, but I'm a believer of applying common sense in these situations.

encoding/descaling.md
```diff
@@ -304,39 +312,48 @@ and `get_w` from `vsutil` to calculate the width)[^1]:
...
-descale_c = core.descale.Debicubic(Y, width, height, b=0, c=1).resize.Bicubic(1920, 1080, filter_param_a=0, filter_param_b=1)            # Sharp Bicubic
-descale_d = core.descale.Debicubic(Y, width, height, b=1, c=0).resize.Bicubic(1920, 1080, filter_param_a=1, filter_param_b=0)            # B-Spline
-descale_e = core.descale.Debicubic(Y, width, height, b=0, c=1/2).resize.Bicubic(1920, 1080, filter_param_a=0, filter_param_b=1/2)        # Catmull-rom
-descale_f = core.descale.Despline36(Y, width, height).resize.Spline36(1920, 1080)                                                        # Spline36
-
-descaled = kgf.join([descale_a, U, V])
-descaled.set_output()
+# Bilinear
+descale_a = core.descale.Debilinear(y, width, height).resize.Bilinear(1920, 1080)
+descale_a = kgf.join([descale_a, U, V])
```
FichteFoll: You need to update all these upper-case references (also applies to the following lines).

FichteFoll (changes requested):
Went through the review again and resolved the addressed feedback. Here's one more, however.

encoding/descaling.md
kageru: This is not an Oxford comma. You always put a comma between two independent clauses joined by a conjunction. “I did A, and I did B” needs a comma. “I did A and B” does not have a comma because “and B” is not an independent sentence.

FichteFoll (approved):

encoding/descaling.md
kageru: nearest neighbor in lowercase

OrangeChannel:
I've sent a [PR](https://github.com/LightArrowsEXE/guide.encode.moe/pull/2) to @LightArrowsEXE , addressing some of kageru's suggestions, and also making sure visually the captions are properly formatted and actually display (due to blank lines not surrounding the attachment line). There are still a few issues to address, namely

- https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/17#discussion_r298801336
- https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/17#discussion_r298801844

Re: the fancy quotes discussion. It is much easier to type `'` , and `"` and is a universal standard according to [the unicode character table](https://unicode-table.com/en/sets/quotation-marks/), but not sure if that's an official source or anything. It seems that the basic quotes are used in all of the other pages, and would require a decent amount of replacing and then probably a mention in the contributing section (these aren't easy to type on a default US keyboard) if we were to change this.

edit: I've also checked that the page and all the links/attachments work properly on a live version of the book, so it should be all good to go once these two issues are addressed.

FichteFoll:
I agree that using fancy quotes is probably too much for a guide like this. I already voiced my specitcism, I believe. Added a review to the mentioned PR (https://github.com/LightArrowsEXE/guide.encode.moe/pull/2).

I believe we can merge once those are addressed.

OrangeChannel:
Bringing this discussion back over here, https://github.com/LightArrowsEXE/guide.encode.moe/pull/2#discussion_r298813839, so I can get others' opinions on this.

`but imo it makes sense to have this be a direct hyperlink, assuming the kagefunc library has docs. I wonder if the glossary would do us any good here because it is going to be in backticks most of the time.`

This is definitely a possibility, would likely require taking kagefunc out of the in-line code as hopefully that doesn't count as a glossary use (will test before commiting). Regarding the glossary entry, what should be put under it besides a hyperlink i.e. https://encode.moe/kagefunc ? @kageru any suggestions?

kageru:
I don’t think we need to write a full glossary entry for Python modules or plugins. A reader can follow the hyperlink(s) and read the readme/docs/docstrings of the module, so writing our own description feels like redundant work to me.

FichteFoll:
The idea with the glossary is that it will insert the links (to the glossary entry) automatically as soon as the word is spotted in use. The alternative would be to manually add a link to every occurance of a module name (but with the advantage of putting it in backticks, i.e. `` [`kagefunc`][kagefunc]``.

FichteFoll:
I'll merge this for now. We can decide on script linking later.

PR #19, opened 2019-06-30
Don't encode a lossy track with lossless codecs

stuxcrystal:
I'm not sure how to wrap at this sentence.

### DIFF
--- a/encoding/preparation.md
+++ b/encoding/preparation.md
@@ -339,7 +339,7 @@ And they both have a point.
 
 So, do some ABX testing and decide for yourself.
 
-Encoding a lossy audio-track with a lossless one is a monumentally stupid idea.
+Encoding a lossy audio-track with a lossless codec is a monumentally stupid idea.
 Lossless Audio Codecs are not magic.
 They can't generate infomation that has been lost by lossy audio compression.
 You will still get the same lossy audio track but you will encode it with a lot more bitrate.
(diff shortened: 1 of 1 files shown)


FichteFoll:
This should be worded more strongly, like "However, it is never an option to transcode lossy audio into lossles because you are using more bitrate for the same information. You cannot generate the information that was lost during compression."

encoding/preparation.md
```diff
@@ -339,6 +339,11 @@ And they both have a point.
 
 So, do some ABX testing and decide for yourself.
 
+Be aware though,
+encoding a lossy audio-track with a lossless codec
+will result in the same lossy audio-track.
+It will require a lot more bitrate though.
```
FichteFoll: This should be worded more strongly, like "However, it is never an option to transcode lossy audio into lossles because you are using more bitrate for the same information. You cannot generate the information that was lost during compression."

FichteFoll (changes requested):

stuxcrystal:
A good alternate wording would be:

    It is always an option to convert a lossy audio-track with a lossless codec.
    Just make sure that you tell *all* users that you did so.
    That way people can see your incredible incompetence,
    stupidity or both.

rokweom:
You're repeating what was already written: 
> Lossily compressed files should generally not be reencoded—the proper way to handle them is to remux (i.e. copy) them to the final file.

I don't see the need to add jokes to this guide.
Besides, reencoding lossy audio is not always stupid - for example when you have audio with PAL speed-up, you need to reencode it after slowing it down. That's why I wrote `generally`.

FichteFoll:
I suggested #23 to supersede this PR.

PR #23, opened 2019-07-03
Reemphasize that encoding lossy to lossless is useless

FichteFoll:
I didn't like the wording suggested in #19 and @rokweom rightfully mentioned that we'd be repeating ourselves, so I propose a different solution that clarifies why you shouldn't reencode in the initial section and later references it for clarity.

### DIFF
--- a/encoding/preparation.md
+++ b/encoding/preparation.md
@@ -304,9 +304,19 @@ adding it to the system PATH is recommeded.
 When working with WEB and TV sources,
 you will most likely have only lossy audio available.
 The most common codecs here are AC-3, E-AC-3 and AAC.
-Lossily compressed files should generally
-not be reencoded—the proper way to handle them
+**Lossily compressed audio
+should generally not be reencoded**—the proper way to handle them
 is to remux (i.e. copy) them to the final file.
+Reencoding lossy audio into a different lossy audio codec
+cannot improve the audio quality
+for barely improved compression
+(in relation and assuming a lower bitrate),
+while encoding into a lossless audio codec
+is just wasting bandwidth.
+Your source file will always have the most data available
+and using a lossless codec for that
+can not generate data lost from the previous compression
+from thin air.
 
 [ffmpeg]: https://www.ffmpeg.org/download.html
 
@@ -370,6 +380,10 @@ be audible anyway.
 And they both have a point.
 
 So, do some ABX testing and decide for yourself.
+
+However, remember that you should not reencode
+lossily compressed audio
+into a lossless codec.
 
 
 ## MKVToolNix


encoding/preparation.md
```diff
@@ -275,8 +275,18 @@ adding it to the system PATH is recommeded.
 When working with WEB and TV sources,
 you will most likely have only lossy audio available.
 The most common codecs here are AC-3, E-AC-3 and AAC.
-Lossily compressed files should generally not be reencoded—the proper way to handle them
+**Lossily compressed audio
+should generally not be reencoded**—the proper way to handle them
 is to remux (i.e. copy) them to the final file.
+Reencoding lossy audio into a different lossy audio codec
+degrades the audio quality
```
rokweom: It doesn't, really. If you take a high quality lossy file and reencode it again in high quality, there won't be any audible difference.
https://bernholdtech.blogspot.com/2013/03/Nine-different-audio-encoders-100-pass-recompression-test.html
As somebody on #encoding-autism rightfully mentioned, this test mostly tests how immune a codec is to its own artifacts, but it still shows that even mulitple reencodings won't hurt.

rokweom:
I feel like this is not really needed. If anything, we should make a separate section for explaining the difference between lossless and lossy, and explain why you should not encode from lossy to lossless there.

encoding/preparation.md
FichteFoll: I specifically mentioned using a different audio codec, so that test is barely relevant for my argument. Reencoding using the same codec with lower bitrate isn't mentioned anywhere.

However, the point needs to be made that reencoding 140kbps eac3 into 140kbps aac doesn't help the slightest besides maybe hardware compatibility (I don't actually know the hardware support of eac3).

FichteFoll:
Rebased and changed the wording slightly.

encoding/preparation.md
rokweom: I'm not saying reencoding with the same codec should be mentioned. I'm just saying that reencoding lossy audio isn't as terribly harmful as many picture it.

And I stand by my previous comment - we should make a separate section for explaining the differences between lossless and lossy. What would you say if I rewrote the "Lossless or lossy?" section? I would explain the differences, what you should and shouldn't do, etc. there.

FichteFoll: Have you checked out my updated wording? It's a bit hard to spot due to the rebase, but it's [here](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/compare/e672aa4c08ffeb9bf99567fbf5590953fcabe7f2..96bdf224254192a61f885b50537409e70dfa4c47#diff-d32a06396d1dd04f5fa407edca2e9180R311).

If think that's okay, I'd like to merge this PR for now, but you may of course also work on the lossless or lossy section. I don't have any particular goals for that myself, though.

PR #25, opened 2019-07-14
 Encoding/preperation: Mention VS Multi-Viewer and add info about VSDB 

Lypheo:
Title says it all.


encoding/preparation.md
```diff
@@ -109,19 +109,22 @@ or [ask for help][Discord].
...
 These are usually more specific than the universally usable core plugins
 or they are collections of wrappers and functions.
 A (non-exhaustive) list of plugins and scripts is available in the
-[official documentation][vs-plugins].
-Alternatively, some individuals in the community are creating archives
-containing the most recent version of useful plugins.
-[eXmendiC's encode pack][ex-encode-pack] is an example of this
-(the lite version should suffice).
+[official documentation][vs-plugins]. Additionally, an extensive database of VapourSynth plugins and scripts is available at [VSDB][vsdb]. 
+VSDB also offers  a GUI for [vsrepo][vsrepo], VapourSynth’s official package manager,
```
FichteFoll: Double space here.

encoding/preparation.md
```diff
@@ -162,8 +165,21 @@ VapourSynth-Script.
...
+     Multi-Viewer is a very simple and elementary previewing tool. 
+     While the text editing is vastly inferior to VSEdit’s, 
+     and the preview lacks a lot of VSEdit’s advanced features,
+     its tab-based previewing functionality easily outclasses VSEdit’s single preview window, 
+     because it makes comparisons between different script versions a lot more convenient and efficient.
+     
+     In short: very useful for thorough comparisons and filter fine-tuning, not so much for everything else.
+     
+     ![The editing window](images/VS-Multi-Viewer_Preview.png)
+     ![the preview window](images/VS-Multi-Viewer_Editor.png)
```
FichteFoll: The title text is used for the figure caption, so it should start with an upper-case letter. As such, mentioning the application's name would also be useful for clarification.

Images need to be surrounded by blank lines.

FichteFoll (commented):
Please add a couple line breaks in the new paragaphs for semantic line feeds. Those should not only be added after each sentence but also within structural parts of a sentence.

Lypheo:
Anything else?

encoding/preparation.md
```diff
@@ -162,8 +173,23 @@ VapourSynth-Script.
...
+     and the preview lacks a lot of VSEdit’s advanced features,
+     its tab-based previewing functionality easily outclasses VSEdit’s single preview window, 
+     because it makes comparisons between different script versions 
+     a lot more convenient and efficient.
+     
+     In short: very useful for thorough comparisons and filter fine-tuning, 
+     not so much for everything else.
+     
+     ![VS Multi-Viewer’s editing window.](images/VS-Multi-Viewer_Preview.png)
+     ![VS Multi-Viewer’s preview window.](images/VS-Multi-Viewer_Editor.png)
```
FichteFoll: Just add a newline before the second image and we should be good.

FichteFoll (changes requested):

encoding/preparation.md
OrangeChannel: This is kind of interesting, by having those lines indented 5 spaces instead of the 4 with the rest of the list items, they don't display a caption ~~and also break the rendered numbering. AvsPmod here becomes item `1.` again, even though it's 4 in the md.~~

Apparently this is because of a blank line, not the indentation.

OrangeChannel:
@Lypheo here's a patch for formatting and some un-related fixes to the preparation page that can be combined here. 

https://gist.github.com/OrangeChannel/662cfe1902dae3c7d5782b9c32dc84e4

Lypheo:
Done.

FichteFoll:
Thanks. Someone should probably check whether embedding the images within an ordered list actually works.

OrangeChannel:
Yea, I test the PRs on a local version of the site before they get merged, and the last commit fixed all the formatting issues (afaict). Works on mobile too.

FichteFoll:
Thanks for checking. I'm too lazy to clone and build locally myself. :sweat_smile:

I also just realized we already have images within the same ordered list for the other editors.

PR #27, opened 2019-07-16
Explain Maximum/Minimum and add more detailed usage example

Lypheo:
This is more of a tentative suggestion than a PR for now: I replaced the old example because it used a private script. The new example, however, includes a bunch of images to make the filtering decisions more understandable and illustrate the effect of various filters. I suspect it may clutter the page and may not be of much benefit, though, so please tell me what you think. Besides, I don’t even have all that much encoding experience myself, so don’t hesitate to call me the idiot that I am if I wrote something stupid. ヽ( ﾟヮ・)ノ


encoding/masking-limiting-etc.md
```diff
@@ -22,14 +22,57 @@ This article will cover:
...
 [**std.Minimum/std.Maximum**](http://www.vapoursynth.com/doc/functions/minimum_maximum.html)
 
-TODO
+The Minimum/Maximum operations replace each pixel
+with the smallest/biggest value in its 3x3 neighbourhood. 
+The 3x3 neighbourhood of a pixel
+are the 8 pixels directly adjacent to the pixel in question
+plus the pixel itself.
+To illustrate:
+![3x3 neighborhood](images/3x3.png)
```
FichteFoll: This would probably be better as a figure surrounded by blank lines and the caption "Illustration of the 3x3 neighborhood" without referring to it with a colon.

encoding/masking-limiting-etc.md
```diff
@@ -22,14 +22,57 @@ This article will cover:
...
+with the brightest (Maximum) or darkest (Minimum) pixel in that neighbourhood.
+
+Maximum generally expands/grows a mask
+because all black pixels adjacent to white edges will be turned white,
+whereas Minimum generally shrinks the mask
+because all white pixels bordering on black ones will be turned black.
+
+See the next section for usage examples.
+
+Side note:
```
FichteFoll, line 61: I wonder if gitbook's markdown has something special for use cases like this (e.g. an info box or similar), but I cba to look for it now.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
 and limits it to the area affected by a line-darkening script.
 The main mask has no name and is simply dhhmask(mode=3).
 
-Example: Build a simple dehalo mask
+For more information about edgemasks, see [here](https://kageru.moe/blog/article/edgemasks).
```
FichteFoll: It's almost never wise to use `here` as the link text. The actual link should also be specified after the paragraph as in-text linking is reserved for links within the same page.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+
+#### Example: Build a simple dehalo mask
+
+Suppose you want to remove these halos:
+
+![source](images/halos.png)
```
FichteFoll: Since the title will be used as the image caption, something more descriptive would be desireable. Same below.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+
+![halos](images/src0.png)
+
+(Note that the images shown in your browser are likely resized poorly; 
+you can compare them at full size [here][comparison].)
```
FichteFoll: This might actually be one of those cases where `here` could be used, but I still prefer a solution without it. Also, comparison is too general of a name. Make this "halo-comparison" and move the url definition closer to its usage, unless it's used multiple times.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+In order to do that,
+we first need to generate an edgemask.
+In this example,
+we’ll use the built-in Sobel function.
+After generating the edge mask, we extract the luma plane:
+
+```py
+mask = core.std.Sobel(src, 0)
+luma = core.std.ShufflePlanes(mask, 0, colorfamily=vs.GRAY)
+
```
FichteFoll: Unnecessary blank line here.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+```
+
+![mask luma](images/luma0.png)
```
FichteFoll: better caption

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+
+Next, we expand the mask twice, so that it covers the halos. 
+``vsutil.iterate`` is a function in [vsutil][vsutil iterate]
```
FichteFoll: If you want to refer to the iterate function in vsutil, the hyperlink needs to span `function in vsutil` as otherwise you'd expect to be referred to vsutil's main docs/page.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+which applies the specified filter a specified number of times to a clip –
```
FichteFoll: Although I personally don't like it, we use `—` without surrounding spaces since this is the standard in English. Unfortunately, the nature of markdown then requires us to not have a line break near `—` because it would be transformed into a space otherwise.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
 
 ```py
-#black line mask
-mask = zz.dhhmask(src)
+mask_outer = vsutil.iterate(luma, core.std.Maximum, 2)
+```
 
-#grow mask to cover halos
-mask_outer = kgf.iterate(mask, core.std.Maximum, 3)
+![mask_outer](images/mask_outer0.png)
```
FichteFoll, line 179: Needs a blank line after this and a caption.

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+Now we shrink the expanded clip back
+to cover only the lineart. 
+Applying ``std.Minimum`` twice
+would shrink it back to the edge mask’s original size, 
+but since the edge mask covers part of the halos too, 
+we need to erode it a little further. 
+
+The reason we used ``mask_outer`` as the basis and shrank it thrice,
```
FichteFoll: Why is this in past tense?

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+preventing detail loss. 
 
-#extend a bit and shrink back to exclude close, adjacent lines
-mask_inner = core.std.Maximum(mask_outer)
-mask_inner = kgf.iterate(mask_inner, core.std.Minimum ,4)
+```py
+mask_inner = vsutil.iterate(mask_outer, core.std.Minimum, 3)
+```
 
+![mask_inner](images/mask_inner0.png)
```
FichteFoll, line 201: needs blank line and caption

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+Now we substract the outer mask covering the halos and the lineart from the inner mask covering only the lineart. 
+This yields a mask covering only the halos,
+which is what we originally wanted:
+
+```py
 halos = core.std.Expr([mask_outer, mask_inner], 'x y -')
 ```
 
+![halos](images/halos0.png)
```
FichteFoll, line 211: blank line and caption

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+Next, we do the actual dehaloing:
+
+```py
+dehalo = hf.DeHalo_alpha(src)
+```
+
+![dehalo](images/dh0.png)
```
FichteFoll, line 219: blank line and caption

encoding/masking-limiting-etc.md
```diff
@@ -83,22 +126,107 @@ Internally it uses a mask called a Camembert to generate a larger mask
...
+leaving the lineart mostly untouched:
+
+```py
+masked_dehalo = core.std.MaskedMerge(src, dehalo, halos)
+```
+
+![voilá: masked_dehalo](images/dehalod0.png)
+
+[comparison]: https://slowpics.org/comparison/96cbeca4-b4be-4dfc-82b1-631bbc85cdb0
+[dehalo_alpha]: https://github.com/HomeOfVapourSynthEvolution/havsfunc/blob/master/havsfunc.py#L393
```
FichteFoll: Linking to lines in master branches may lead to the line numbers changing as the code is updated. Although permalinking older revisions might point to outdated versions, it's still better as the reference will continue to work properly when the source is updated.

You can retrieve a permalink with the latest revision of a file by pressing `y` while browsing Github.

FichteFoll (changes requested):
I'm not an encoder, but the contents look good. Thanks.

I left some comments mostly regarding formatting and other stylistic things.

encoding/masking-limiting-etc.md
OrangeChannel: The new gitbook spec does lol <https://docs.gitbook.com/content-editing/rich-content#hints-and-callouts>.
The only way to do something like this would be with HTML.

OrangeChannel: image
image
image
Something like this, maybe?
It's responsive to the theme changing.
Link formatting would have to change, making the file look pretty ugly.

OrangeChannel:
Don't worry too much about whitespace-related formatting yet, @Lypheo. Once the content gets approved I'll go through everything with a running version of the book. The other suggestions still hold.

encoding/masking-limiting-etc.md
FichteFoll: Not unexpected that the new spec has it. The old might as well, but docs are kinda lost now, I guess? Maybe waybackmachiene or archive.org has something.

Your alternative looks decent enough, though.

encoding/masking-limiting-etc.md
FichteFoll: Uh, you changed this for `vsutil iterate` but not for `dehalo_alpha`.

Lypheo: Oops. Fixed.

PR #28, opened 2019-07-19
Updates to contribution guidelines, fixes

OrangeChannel:
Tested working on browser. Haven't tested on mobile but doubt anything is different.

Also tried playing around with the new gitbook format and well... the "GitHub integration" sure is [something](https://github.com/OrangeChannel/fansubbing-guide/commit/e5205b8de2340cc4a5da87dbea195e9e69525ba3). Doesn't allow for editing from non-organization members (2 people max for free) and it replaces all the markdown syntax with garbage. Plug-ins are mostly gone, there's no way to use footnotes, and glossary is gone. A lot of formatting is also gone, and theme-ing is pay-to-win.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -357,7 +357,7 @@ Line breaks **cannot** be used.
 
 ---
 
-[^1]: The new gitbook spec is very different than the version this book is using. Almost none one of the information from [gitbook's new website][new-gitbook] applies.
+[^1]: The new gitbook spec is very different than the version this book is using. Almost none of the information from [gitbook's new website][new-gitbook] applies.
 
 [^2]: This is different from how github.com's markdown preview behaves.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,14 @@ and automatically reloaded.
 [node.js]: https://nodejs.org/
 
 
+### Adding a New Page
+
+In order for your page to be accessible,
+you need to add it to the `SUMMARY.md` file.
+The title used there will be used in the navigation bar,
+so keep it short.
+
+
 ### TODO
 
 Various sections are still under construction.
@@ -188,7 +196,9 @@ parenthesized sentences (not words),
 or new items in a long list
 (such as the one you are reading right now).
 
-[Semantic Linefeeds]: http://rhodesmill.org/brandon/2012/one-sentence-per-line/
+[Footnotes](#footnotes) are an exception and must be written all on one line.
+
+[Semantic Linefeeds]: https://rhodesmill.org/brandon/2012/one-sentence-per-line/
 
 
 ### Indentation
@@ -231,8 +241,10 @@ and **one** after.
 Headings immediately following their parent heading
 only need one blank line in-between.
 
-Separate text from end-of-section hyperlink lists with **one** blank line
-before the list. See [below](#Hyperlinking) for more information.
+Additionally, separate text from end-of-section hyperlink lists
+with **one** blank line before the list.
+For image embeds,
+there should be a blank line on each side.
 
 
 ### Hyperlinking
@@ -265,11 +277,6 @@ you will want multiple lines at the end of the header section.
 …
 ```
 
-If you are linking to a section on the same page,
-`[section name](#header)` is allowed in-line.
-An example of this is [the hyperlink section you are reading](#Hyperlinking).
-In markdown, this is simply `[the hyperlink section you are reading](#Hyperlinking)`.
-
 For relative links (links to other pages,
 images,
 or files within this repository),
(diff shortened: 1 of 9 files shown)


FichteFoll:
Yep, the new gitbook format is a no-go. If we ever run into something we can't solve with gitbook but would like to have regardless, we'd need to look into an alternative.

I'll review this later.

OrangeChannel:
CSS changes in a07dd1b :

before
Before

after
After

OrangeChannel:
Need @stuxcrystal to confirm these two new pages are fine, as they use his information, and are mostly copied from yuuno's site.

stuxcrystal:
Perfect. Just make sure this page is accessible by at most two clicks from any page on the site.

Which, I think, is given here.

CONTRIBUTING.md
```diff
@@ -53,14 +53,10 @@ why it only applies to a specific language.
...
 
 This guide is written in [Markdown][]
-and uses [gitbook][]'s [toolchain][] to compile
+and uses legacy gitbook's [toolchain][] to compile
 the static HTML pages.
 Gitbook uses the [Github Flavoured Markdown][GFM] (GFM) variant.
 
-Documentation for gitbook's toolchain (including gitbook markdown sytax)
-can be found here:
-<https://toolchain.gitbook.com/>
```
FichteFoll, line 62: Ah, they nuked it. Too bad.

CONTRIBUTING.md
```diff
@@ -294,9 +330,35 @@ For a full document citation,
...
+Examples of this can be found on the
+[Aegisub and Other Tools](typesetting/aegisub.md) page.
+
+
+### Footnotes
+
+Footnotes can be used for information that would interrupt
+the flow or purpose of a paragraph,
+but may still be of interest.
+They are created with `[^#]` in-text,
```
FichteFoll, line 350: An explanation that `#` should be replaced by a number is necessary here.

typesetting/aegisub.md
```diff
@@ -28,12 +28,10 @@ Here is a list of tools you will want to download:
           - [MainType][]
           - [Suitcase Fusion][]
   - Software for motion-tracking
-      - Option 1: [Mocha Pro standalone app][]
+      - [Mocha Pro standalone app][]
         - Look for it on *certain websites*.
         - On Windows, Mocha **requires** [Quicktime][] to be installed.
           More information can be found [here][quicktimeFAQ].
-      - Option 2: [Blender][][^1]
```
FichteFoll, line 35: While Blender is a valid option, I hardly know anyone who knows how to use it, let along be willing to add it to this guide. I guess less is more in this case?

overview/roles.md
```diff
@@ -90,20 +90,20 @@ presentation of translated text on-screen. These are generally called *signs*.
 
 For example, given this scene and a translation of "Adachi Fourth Public High School"…
 
-![\[DameDesuYo\] Eromanga-sensei - 01 (1920x1080 10bit AAC) \[05CB518E\].mkv\_snapshot\_03.11\_\[2017.08.18\_21.14.55\].jpg](images/cnvimage100.png)
+![[DameDesuYo] Eromanga-sensei - 01 (1920x1080 10bit AAC) [05CB518E].mkv_snapshot_03.11_[2017.08.18_21.14.55].jpg](images/cnvimage100.png)
```
FichteFoll, line 93: Don't these underscores cause text to be italicized?

FichteFoll (commented):
Very good changes. Thanks for these.

Please just see a couple of my comments.

FichteFoll:
>Perfect. Just make sure this page is accessible by at most two clicks from any page on the site.

It will in fact always be exactly one click.

typesetting/aegisub.md
OrangeChannel: I was going to bother explaining it, updating some of the information [here](https://subarashii-no-fansub.github.io/Subbing-Tutorial/Tracking-Motion/), but until v2.80 comes out, I don't think it's worth using. If 2.80 makes it much easier to do (with an actual decent GUI this time) I could add it back. I recall the [export script](https://raw.githubusercontent.com/Subarashii-no-Fansub/AAE-Export/master/aae-export.py) not working with a beta version of the new Blender too, but I could be thinking of another script.

overview/roles.md
OrangeChannel: Oddly enough, no they don't. It seems the caption plug-in doesn't recognize markdown formatting, and just italicizes everything. (The asterisks are just for an example, and aren't escaped)

image

OrangeChannel: This is on the descaling page for example, no formatting.

image

FichteFoll: Thanks for checking.

CONTRIBUTING.md
FichteFoll: Addressed in 6c39f575d97f3431892d1b2d33b28d4a0d30edb4.

PR #29, opened 2019-07-21
Update impressum.md

stuxcrystal:
Most automated impressum-generators on the internet require a link to be placed back to them.

Since I certainly did not pay for the impressum generator, they want to have a link placed denoting where I got this thing.


OrangeChannel:
@FichteFoll can you edit this commit to change `[imp-tl][Englisch-Übersetzungsdienst translate-24h]` to `[Englisch-Übersetzungsdienst translate-24h][imp-tl]` so the link works?

FichteFoll:
Good catch. Fixed in 3149028bfd707d9e3ad80840cebf222e543cea20.

PR #31, opened 2019-08-04
Fixes and cleanup for masking-limiting-etc.md

OrangeChannel:
This is what the TODO button thingy looks like. It links the same way the `EDIT THIS PAGE` button at the top does.

Screenshot_20190804_205929


FichteFoll:
Thanks for these changes. The TODO buttons seem a bit overengineered imo, but surely won't hurt. Doesn't need to be added everywhere, however.

PR #33, opened 2019-08-24
Explain what a mask even is

Lypheo:
Moved the dehalo example down because it didn’t fit well at the top.

### DIFF
--- a/encoding/masking-limiting-etc.md
+++ b/encoding/masking-limiting-etc.md
@@ -1,7 +1,7 @@
 # Masking, Limiting, and Related Functions
 
 There are filters
-which changes the video in various ways,
+which change the video in various ways,
 and then there are ways to change the filtering itself.
 There are likely hundreds of different techniques at your disposal
 for various situations,
@@ -20,8 +20,8 @@ This article will cover:
 
 ## Masking
 
-Masking refers to a broad set of techniques used to merge multiple clips,
-usually one filtered clip merged with a source clip
+Masking refers to a broad set of techniques used to merge multiple clips.
+Usually one filtered clip is merged with a source clip
 according to an overlay mask clip.
 A mask clip specifies the weight for each individual pixel
 according to which the two clips are merged;
@@ -57,24 +57,25 @@ or `GRAYS` (single precision floating point).
 This is the main function for masking
 that performs the actual merging.
 It takes three clips as input:
-clipa, clipb and one mask clip.
+two source clips and one mask clip.
 The output will be a convex combination of the input clips,
-where the weights are given by the brightness of the mask clip,
-like this
-(the variables refer to the value of the same pixel in each clip):
+where the weights are given by the brightness of the mask clip.
+The following formula
+describes these internals for each pixel:
 
 ```py
-output = clipa * (MAX_VALUE - mask) + clipb * mask # in 8 bpp for example, MAX_VALUE would be 255
+# in 8 bpp, MAX_VALUE would be 255
+output = clipa * (MAX_VALUE - mask) + clipb * mask 
 ```
 
 In simpler terms:
 for brighter areas in the mask,
-the output will come from clipb,
+the output will come from `clipb`,
 and for the dark areas,
-it’ll come from clipa.
-Grey areas result in an average of clipa and clipb.
+it’ll come from `clipa`.
+Grey areas result in an average of `clipa` and `clipb`.
 
-If ``premultiplied`` is set to True,
+If `premultiplied` is set to True,
 the equation changes as follows:
 
 ```py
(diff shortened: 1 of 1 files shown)


encoding/masking-limiting-etc.md
```diff
@@ -22,9 +22,28 @@ This article will cover:
...
+e.g. with an edgemask,
+and merging the filtered clip with the unfiltered clip
+according to the mask,
+such that the masked areas are taken from the unfiltered clip,
+and the unmasked areas are taken from the filtered clip.
+In effect,
+this applies the filtering only to the unmasked areas of the clip,
+leaving the masked details/edges intact.
+(In this context,
+“masked” means that the pixels in question are white in the mask clip.)
```
FichteFoll: This parenthesized explanation doesn't really help with understanding the internals because it's too unspecific. Either be more specific on how it works or leave it out entirely since the more detailed explanation is in line 55.

encoding/masking-limiting-etc.md
```diff
@@ -33,8 +52,47 @@ and one of these formats:
 `GRAY16` (16 bits integer),
 or `GRAYS` (single precision floating point).
 
-Vapoursynth includes some basic tools for manipulating masks as well:
+#### [std.MaskedMerge][]
```
FichteFoll, line 53: Eh, going from from h1 to h4 is suboptimal, but using h2 for a filter seems superfluous.

encoding/masking-limiting-etc.md
```diff
@@ -33,8 +52,47 @@ and one of these formats:
...
+like this
+(the variables refer to the value of the same pixel in each clip):
+
+```py
+output = clipa * (MAX_VALUE - mask) + clipb * mask # in 8 bpp for example, MAX_VALUE would be 255
+```
+
+In simpler terms:
+for brighter areas in the mask,
+the output will come from clipb,
```
FichteFoll: I suggest using inline code blocks when referring to variable names, so `` `clipb` ``. (Actually I went ahead and did that.)

encoding/masking-limiting-etc.md
```diff
@@ -138,6 +196,34 @@ see [kageru's blog post][].
 
 [kageru's blog post]: https://kageru.moe/blog/article/edgemasks
 
+I would also lump the range mask
```
FichteFoll: With how many different authors we have now, using `I` seems odd.
That's just for futore thought, however, since you didn't change that in this PR.

Lypheo: Perhaps put the author's name (Zastin, I think) in parantheses after the “I”?

encoding/masking-limiting-etc.md
Lypheo: So what do you suggest, then? 🤔

FichteFoll: If I had a suggestion, I would have already given it. 😄 

I say leave it as h4 for now, because it's like that for the rest of the file already and looks good enough. Until we figure out something better, this will most certainly do. h3 is still too prominent in the default style. You can resolve this.

encoding/masking-limiting-etc.md
FichteFoll: Either that, or rephrase to be less opinionated and more generally applicable, I say. Or, well, just remove the personal apsect of the message and do a well-informed best practice suggestion.

FichteFoll:
I suggest explaining std.Merge before std.MaskedMerge because the latter is the more complicated form of the former, thus allowing to explain the general concept for the Merge function and then explaining how a mask clip helps.

Lypheo:
> I suggest explaining std.Merge before std.MaskedMerge because the latter is the more complicated form of the former, thus allowing to explain the general concept for the Merge function and then explaining how a mask clip helps.

I’m not sure if that order would make more sense, because I deliberately placed MaskedMerge right after the general description of masking in order to keep the explanation cohesive. Putting Merge above it doesn’t help the flow much IMO.

FichteFoll:
Hm, but what about explaining std.Merge before masking concepts? I mean you need the concept of merging two clips before you can merge them selectively.

encoding/masking-limiting-etc.md
OrangeChannel: I changed around the wording on all the uses of `I`, let me know what you guys think.

OrangeChannel:
> Hm, but what about explaining std.Merge before masking concepts? I mean you need the concept of merging two clips before you can merge them selectively.

The way it's organized right now, this would require moving the `Single and multi-clip adjustments with std.Expr and friends` section above the `Masking` one.

FichteFoll:
Well, we can revisit this later if someone sees the need. For now, let's get this shipped. 🚀

PR #34, opened 2019-09-01
Scenefiltering.md update

OrangeChannel:
Slight rewrite for clarity, and I updated the code-blocks to better show the filterchaining/rfs IMO. 
Also added some info for shortcuts using VSEdit and [this bookmark script](https://gist.github.com/OrangeChannel/b9666b3650a3448589069d25dd6a394c) I wrote.

### DIFF
--- a/encoding/scenefiltering.md
+++ b/encoding/scenefiltering.md
@@ -146,8 +146,30 @@ This step can take anywhere from a few minutes to hours, depending on
 the encoder and the source. Most of the time, the same filters can be
 reused every episode with some minor changes here and there.
 
-TODO: Talk about which editor to use and how (bookmarks, comparisons,
-etc.)
+
+### Editor shortcuts / tips
+
+If using VSEdit as your [editor](preparation.md#the-editor),
+it can be helpful to use the
+built-in bookmark functionality
+to find the frame ranges of each scene.
+There is a [small script][vsbookmark] that can generate
+these bookmarks from your clip inside of VSEdit.
+
+```py
+# Editing a script called 'example01.vpy'
+import ...
+from vsbookmark import generate
+
+generate(clip, 'example01')
+clip.set_output()
+```
+
+When previewing your clip,
+there will now be bookmarks generated on the timeline
+allowing you to skip to the next scene using the GUI buttons.
+
+---
 
 Now you might ask, "Why did I have to create base filters for
 everything?" The answer is that these base filters allow other filters
@@ -170,3 +192,5 @@ weak, faster filter multiple times.
 
 That should be all you need to know
 about scenefiltering.
+
+[vsbookmark]: https://gist.github.com/OrangeChannel/b9666b3650a3448589069d25dd6a394c
--- a/encoding/scenefiltering.md
+++ b/encoding/scenefiltering.md
@@ -28,7 +28,7 @@ your video source:
 import vapoursynth as vs  # this can look different based on your editor
 core = vs.get_core()      # the following uses VSEdit's format
 
-src = core.lsmas.LWLibavSource("source.mkv")
+src = core.lsmas.LWLibavSource("source.m2ts")
 ```
 
 Next, you need to choose what filtering will be done to the entire clip.
@@ -60,16 +60,17 @@ time. At this point, your script should look something like this:
 import vapoursynth as vs
 core = vs.get_core()
 
-src = core.lsmas.LWLibavSource("source.mkv")
-filtered = default_filtering(src)
+src = core.lsmas.LWLibavSource("source.m2ts")
+filtered = core.resize.Bilinear(src, width=1280, height=720)
 
-light_denoise = someDenoiseFilter(filtered)
-heavy_denoise = someOtherDenoiseFilter(filtered)
+light_denoise = some_denoise_filter(filtered)
+heavy_denoise = some_other_denoise_filter(filtered)
 
 aa = antialiasing(filtered)
 
-light_deband = deband1(filtered)
-med_deband   = deband2(filtered)
+default_deband = deband(filtered)
+light_deband   = deband1(filtered)
+medium_deband  = deband2(filtered)
 ```
 
 
@@ -89,21 +90,19 @@ import vapoursynth as vs
 import fvsfunc as fvf
 core = vs.get_core()
 
-src = core.lsmas.LWLibavSource("source.mkv")
-filtered = defaultFiltering(src)
+src = core.lsmas.LWLibavSource("source.m2ts")
+filtered = core.resize.Bilinear(src, width=1280, height=720)
 
 ### Denoising
-
-light_denoise   = someDenoiseFilter(filtered)
-heavy_denoise   = someOtherDenoiseFilter(filtered)
-heavier_denoise = someStrongerDenoiseFilter(filtered)
+light_denoise   = some_denoise_filter(filtered)
+heavy_denoise   = some_other_denoise_filter(filtered)
+heavier_denoise = some_stronger_denoise_filter(filtered)
 
 denoised = fvf.rfs(filtered, light_denoise, mappings="")
 denoised = fvf.rfs(denoised, heavy_denoise, mappings="")
 denoised = fvf.rfs(denoised, heavier_denoise, mappings="")
 
 ### Anti-aliasing
-
 eedi2_aa  = eedi2_aa_filter(denoised)
 nnedi3_aa = nnedi3_aa_filter(denoised)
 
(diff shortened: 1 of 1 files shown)


encoding/scenefiltering.md
```diff
@@ -21,11 +21,13 @@ fix both of these issues by using strong filtering over the entire
...
 do not want. This is where scenefiltering comes in.
 
-As always, you start by importing the Vapoursynth module and loading
+As always, you start by importing the VapourSynth module and loading
 your video source:
 
 ```py
-from vapoursynth import core
+import vapoursynth as vs  # this can look different based on your editor
+core = vs.get_core()      # the following uses VSEdit's format
```
FichteFoll: I'm not sure I'm on top of the VS changes, but wasn't `from vapoursynth import core` the new recommended way? Or is it vice versa?

Or is this what you mean with "VSEdit's format"?

encoding/scenefiltering.md
```diff
@@ -47,37 +49,66 @@ part well, you will save yourself a lot of time later on, so take your
...
+import vapoursynth as vs
+core = vs.get_core()
+
 src = core.lsmas.LWLibavSource("source.mkv")
 filtered = default_filtering(src)
-light_denoise = some_denoise_filter(filtered)
-heavy_denoise = some_other_denoise_filter(filtered)
+
+light_denoise = someDenoiseFilter(filtered)
+heavy_denoise = someOtherDenoiseFilter(filtered)
```
FichteFoll: Why change these to camelCase but leave `default_filtering` as-is? Actually, why do it in the first place? We're not imitating some pseudo plugin-based filter like this, so we might as well stick to Python conventions.

encoding/scenefiltering.md
```diff
@@ -115,20 +146,43 @@ This step can take anywhere from a few minutes to hours, depending on
...
+to find the frame ranges of each scene.
+There is a [small script][vsbookmark] that can generate
+these bookmarks from your clip inside of VSEdit.
+
+```py
+# Editing a script called 'example01.vpy'
+import ...
+from vsbookmark import generate
+
+generate(clip, 'example01')
```
FichteFoll, line 199: Maybe also add a function to concert conventional keyframe files into VSEdit's bookmark format?

encoding/scenefiltering.md
```diff
@@ -115,20 +146,43 @@ This step can take anywhere from a few minutes to hours, depending on
...
+clip.set_output()
+```
+
+When previewing your clip,
+there will now be bookmarks generated on the timeline
+allowing you to skip to the next scene using the GUI buttons.
+
+---
 
 Now you might ask, "Why did I have to create base filters for
```
FichteFoll: I feel like this section is out of place after the "Editor shortcuts" section.

encoding/scenefiltering.md
```diff
@@ -47,37 +49,66 @@ part well, you will save yourself a lot of time later on, so take your
...
 Once you've done all of that, you're done with filtering your source—at
 least for the most part. Now all you need to do is add
 `ReplaceFramesSimple` calls. For this, you require the
 plugin [RemapFrames](https://github.com/Irrational-Encoding-Wizardry/Vapoursynth-RemapFrames/releases) or
 the native Python version in
 [fvsfunc](https://github.com/Irrational-Encoding-Wizardry/fvsfunc/blob/master/fvsfunc.py).
-`Rfs` is a shorthand for `ReplaceFramesSimple` and is much easier to
-type, so you should use that.
+`Rfs` is a shorthand for `ReplaceFramesSimple`
+and fvsfunc has the alias `rfs`.
```
FichteFoll, line 92: An explanation about how these two differ could be useful? Or are there none?

FichteFoll (commented):
Yeah, I reviewed this like 6 hours ago, then did *something else*, then remembered I wasn't finished. Submitting the current state anyway because I don't think there's much else to worry about.

encoding/scenefiltering.md
OrangeChannel: Added a `convert` function for WWXD qp-files and (SC)Xvid keyframes text files.

encoding/scenefiltering.md
OrangeChannel: Yea, these two lines are automatically added when you start a new script in VSEdit. It will likely be what most people see at the top of their script.

encoding/scenefiltering.md
OrangeChannel: Moved the Editor shortcuts section to the bottom and removed horizontal rule.

encoding/scenefiltering.md
```diff
@@ -60,16 +60,17 @@ time. At this point, your script should look something like this:
 import vapoursynth as vs
 core = vs.get_core()
 
-src = core.lsmas.LWLibavSource("source.mkv")
-filtered = default_filtering(src)
+src = core.lsmas.LWLibavSource("source.m2ts")
+filtered = core.resize.Bilinear(src, width=1280, height=720)
```
FichteFoll: I actually find it weird to call the resized variant `filtered` (even though you say it's the "base filtering") when you do debanding, denoising and aa later. Maybe just call it `resized`?

encoding/scenefiltering.md
```diff
@@ -60,16 +60,17 @@ time. At this point, your script should look something like this:
...
+light_denoise = some_denoise_filter(filtered)
+heavy_denoise = some_other_denoise_filter(filtered)
 
 aa = antialiasing(filtered)
 
-light_deband = deband1(filtered)
-med_deband   = deband2(filtered)
+default_deband = deband(filtered)
+light_deband   = deband1(filtered)
+medium_deband  = deband2(filtered)
```
FichteFoll: This is more of a principal thing, but wouldn't you deband the denoised clip? In which case the pipeline becomes more like this:

1. denoise
1. (scenefilter denoising)
1. deband
1. (scenefilter debanding)
1. (re-add noise/grain)

Actually, the next example does it like I propose here, so it's just this example here that needs work.

encoding/scenefiltering.md
FichteFoll: As discussed in #encode-autism, this should become
```
import vapoursyth as vs
core = vs.core
```

encoding/scenefiltering.md
OrangeChannel: Do you mean the fact that I am applying the deband to `filtered` here? Because I was under the impression that denoise, aa, deband is the correct order if you need to aa the clip.

If it's the `filtered` thing, that's because this example is just for creating the base filters. But I should probably change these so they look more similar to the next block.

encoding/scenefiltering.md
OrangeChannel: I'm going to leave the testing to someone else, but I'll add a footnote incase anyone is interested and maybe wants to test it themselves.

encoding/scenefiltering.md
FichteFoll: My main point is that using `filtered` here invalidates the pipeline because when you rfs later you only frames that are either debanded or denoised but not both.

Edit: But I see now that you addressed this in a commit, so this can be resolved.

encoding/scenefiltering.md
```diff
@@ -26,7 +26,7 @@ your video source:
 
 ```py
 import vapoursynth as vs  # this can look different based on your editor
```
FichteFoll, line 28: Can remove the comment as well, imo.

encoding/scenefiltering.md
```diff
@@ -61,16 +61,16 @@ import vapoursynth as vs
...
-filtered = core.resize.Bilinear(src, width=1280, height=720)
+resized = core.resize.Bilinear(src, width=1280, height=720)
 
-light_denoise = some_denoise_filter(filtered)
-heavy_denoise = some_other_denoise_filter(filtered)
+light_denoise = some_denoise_filter(resized)
+heavy_denoise = some_other_denoise_filter(resized)
 
-aa = antialiasing(filtered)
+aa = antialiasing(denoised)
```
FichteFoll, line 71: Honestly, having this skeleton without explaining that we're going to use rfs with these multiple clips later is more misleading that it helps, I think.

I know this page is about scenefiltering and explaining how to structure your code, but at this line the example just falls apart. We might as well not include it at all.

encoding/scenefiltering.md
OrangeChannel: yuuno still uses `%load_ext yuuno` for the VapourSynth import, so I think this should stay.

encoding/scenefiltering.md
OrangeChannel: > # Creating the base filters
> Now that you have your common filtering down, you need to create some base filter chains. Go through some random scenes in your source and write down parts of the filtering that best suits those scenes.

I think it's clear that this part is just for creating exactly what is shown in this block, generic filters that we will use later on specific scenes.

i.e. if some scenes you looked through would be fixed with the func `antialiasing` here, you should write it down to choose what scenes need it later

edit: The section header kind of implies this is only a part of scenefiltering.

also i just realized i somehow edited *your* comment with that ^ line, sorry

FichteFoll: Hm, I guess you are right. Since I only read the diff, I didn't get the whole context on this page as I don't think I ever read it. I think we can make this clearer if you add a `denoised = …` line below the various denoised clips so that it becomes more obvious that something is missing here **and** `denoised` is actually defined. Probably do the samefor `debanded = …` below for consistency.

OrangeChannel:
This should be good to go now, let me know if there's anything else to change/add.

FichteFoll:
Nah, we good now. :+1:

PR #36, opened 2019-09-03
Add new CSS rules for image scaling

OrangeChannel:
- on a full width display, images will be scaled down to stop taking up so much space vertically
  - when hovered over, will return to 100% (of the .page-inner, so actually just <=90em)
  - if the image width is already below 70% of <=90em, it won't be scaled and hover will just add a small shadow behind

Animation preview: https://jsfiddle.net/wk6drxp5/

Currently:
current

---

After:
future
`:hover`
image

---

also adds a test line for darkening entire website behind the picture when hovered over, maybe will be useful for something else

### DIFF
--- a/styles/website.css
+++ b/styles/website.css
@@ -508,16 +508,42 @@ figcaption {
     margin-top: 1px;
 }
 
-/* image hovering shadow */
+/* image hovering shadow/zoom */
 img:hover {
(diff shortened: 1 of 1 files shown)


FichteFoll:
I'm too lazy to build locally, so let's test this in production. :+1:

I also thought that images were a bit too large by default.

Issue #22, opened 2019-07-01
Add example images for video artifacts


FichteFoll:
I was made aware of the fact that we don't have any example image for banding. This needs to be changed.

Also many other artifacts mentioned in https://guide.encode.moe/encoding/video-artifacts.html.

FichteFoll:
We do have an example for banding in the descale page using Akanesasu Shoujo. I suppose we could use the same image in the recognizing artifacts page.

However, we're also missing examples for several other artifacts, so I'll expand this issue to cover those as well.

PR #37, opened 2019-09-16
add Overflow/Underflow to Video Artifacts

MysteryDove:
I added a video artifacts to the video-artifacts called Overflow/Underflow with some pics to show that.

I'm not confident with my English, so please feel free to correct or edit the new part


MysteryDove:
I see #22 and added an example image of banding

encoding/video-artifacts.md
```diff
@@ -232,6 +234,27 @@ or a manual filtering of dehalo\_alpha with dhhmask
 
 [![field-noise.jpg](images/3cnvimage103.png)](https://diff.pics/84URvW5IYSdO/1)
 
+### Underflow/Overflow
+
+While almost all of the anime produced
+use the YUV 8-bit limited range (Y: \[16,235\], UV: \[16,240\]),
```
FichteFoll: I think this could use a footnote with a more detailed explanation as to why this limitation exists. I personally didn't know about this, so I had to look it up *somewhere*, but I'm not satisfied enough with my source to reference it here.

OrangeChannel:
Looks good to me. Thanks @pwinner :+1: .

encoding/video-artifacts.md
```diff
@@ -232,6 +234,29 @@ or a manual filtering of dehalo\_alpha with dhhmask
...
+to check the brightness levels with `hist.Levels()`.
+
+[![Example of underflow (click for comparison)](images/underflow.jpg)](https://slowpics.org/comparison/f125e799-fdff-4c4c-8c9d-707af021bd88)
+
+[![Example of overflow (click for comparison)](images/overflow.jpg)](https://slowpics.org/comparison/6e24ffe9-e068-4f33-b2e7-639031d512f2)
+
+To fix this problem,
+simply use `mvf.Depth(input, fulls=True, fulld=False)`
+or set the "full range" flag on the video,
+so the values can be interpreted accordingly.
```
FichteFoll, line 265: Is there any advantage in scaling under- or overflowing video into the limited range compared to setting the full range flag? I would imagine the limited range being more widely supported by players while reducing the range obviously means reducing the available information in the video. Or what exactly does `mvf.Depth` do to begin with?

Furthermore, is this relevant when encoding with 10-bit?

encoding/video-artifacts.md
```diff
@@ -232,6 +234,29 @@ or a manual filtering of dehalo\_alpha with dhhmask
...
+
+### Underflow/Overflow
+
+While most of the anime produced
+use the YUV 8-bit limited range[^4],
+we occasionally find some videos having the "limited range" flag set
+while containing full range content.
+This often results in oversaturated colors and weird brightness.
+Thus, it is strongly recommended
+to check the brightness levels with `hist.Levels()`.
```
FichteFoll: This should probably be applied to the 8-bit source clip, correct?

FichteFoll (commented):
Thanks for the addition. I reworded some sections and fixed the grammar. I also tried to add a somewhat technical information for this limited range problem that I found after some research (because I didn't really know about this before as well).

Please fact-check.

encoding/video-artifacts.md
MysteryDove: Yes! hist.Levels() only works on 8bit, we should mention that

encoding/video-artifacts.md
MysteryDove: I'm not familiar with the reason why there exists the limitation too.
I think this is related to the conversion between YUV and RGB.
If we need to explain this, we need to explain how the conversion worked, and that may be a long story

encoding/video-artifacts.md
MysteryDove: Limited range have greater compatibility while some players may ignore the "full range" flag and treat the video as a limited one.
Then we should process this problem under 16bit or even float point, there shouldn't be any noticeable loss.
Using `mvf.Depth` is a simple but not such a good way because sometimes these problems are complicated, and you won't expect any precision problems while using it.

encoding/video-artifacts.md
```diff
@@ -263,4 +265,6 @@ the use of `std.Expr()` is encouraged.
 
 [^3]: Blocking may also occur for other reasons other than compression data loss. [Image re-construction with padding][waifu2x238] can cause very similar looking effects, although this is irrelevant for fansubbing source videos.
 
+[^4]: The limited range is used in rec.601 and rec.709 and allows values in \[16,235\] for the Y and \[16,240\] for the U and V planes. This means that Y=16 is considered full black and Y=235 full white, while any values outside of that range are clamped. U and V behave analogously.
```
Lypheo: It’s used in BT.2020/2100 as well. For 10/12 bit, the range is 64-940/256-3760 for luma, and 64-960/256-3840 for chroma.

MysteryDove:
I've added some code examples about using Expr to fix Overflow and Underflow. I think the comment part of the code is not good enough. Please feel free to edit this part to make a better expression.

encoding/video-artifacts.md
FichteFoll: Is bt.2020 only for 10bit and 2100 for 12 bit? Color spaces are probably the areay I have the least expertise in besides quantization, so any suggestions for this part are welcome. Would it be enough to just say this is for 8bit?

encoding/video-artifacts.md
```diff
@@ -256,6 +256,10 @@ so the values can be interpreted accordingly.
 If the range conversion is more complicated,
 e.g. from \[16-255\] to \[16-235\],
 the use of `std.Expr()` is encouraged.
+What's more, if you consider using Expr,
+it's better to do this under 16bit to avoid precision problems.
+
+![](images/overflow_notice.jpg)
```
FichteFoll: The image is hard to understand. What message is it supposed to convey?

encoding/video-artifacts.md
```diff
@@ -256,6 +256,10 @@ so the values can be interpreted accordingly.
 If the range conversion is more complicated,
 e.g. from \[16-255\] to \[16-235\],
 the use of `std.Expr()` is encouraged.
+What's more, if you consider using Expr,
+it's better to do this under 16bit to avoid precision problems.
```
FichteFoll: Does 16 bit precision use floats?

encoding/video-artifacts.md
```diff
@@ -261,6 +261,17 @@ it's better to do this under 16bit to avoid precision problems.
 
 ![](images/overflow_notice.jpg)
 
+Here's some examples if you want to use `std.Expr()`.
```
FichteFoll: Good of you to add these. I thought whether we should provide them or have the user figure it out by themselves, but I believe at this point it's better to go with an example to prevent avoidable errors.

encoding/video-artifacts.md
```diff
@@ -261,6 +261,17 @@ it's better to do this under 16bit to avoid precision problems.
...
+
+```py
+# We suppose "src16" is a 16bit video clip and we only process Y plane in this example as well as bypass the UV planes.
+# I have to underline that this example is processed under 16bit and I don't recommend to do this at any lower precisions.
```
FichteFoll: Please try not to use "I" in either text or comments. Passive forms are probably the way to go here. I can also do this, if you'd like me to.

encoding/video-artifacts.md
MysteryDove: I think it's hard to let the user figure it out without having a brief introduction about the conversion between limited range and full range. Or we should make another page introducing these mathematic stuff about YUV, RGB  and range.

encoding/video-artifacts.md
MysteryDove: No, the output result with Expr should be in integer.

encoding/video-artifacts.md
MysteryDove: Great! Please go ahead and do this.

encoding/video-artifacts.md
Lypheo: No, they both define UHD television (whereas 709 is for (Full) HD), but 2020 is for SDR and 2100 is for HDR.
>Would it be enough to just say this is for 8bit?

I suppose. It’s just that the way it’s currently phrased implies limited range YCbCr is some legacy format, which is not the case.

encoding/video-artifacts.md
```diff
@@ -261,6 +261,21 @@ it's better to do this under 16bit to avoid precision problems.
...
+
+Here's some examples if you want to use `std.Expr()`.
+
+```py
+# We suppose "src16" is a 16bit video clip and we only process Y plane in this example as well as bypass the UV planes.
+# I have to underline that this example is processed under 16bit and I don't recommend to do this at any lower precisions.
+# If you want to process UV plane, you need to take care of these linear map expressions because the upper and lower limits in UV planes are different.
+YC = core.std.Expr(src16, expr=["x 255 / 219 * 4096 +",""], format=vs.YUV)   #this makes [0-255] to [16-235] at y plane
+YC = core.std.Expr(src16, expr=["x 4096 - 219 239 / * 4096 +",""], format=vs.YUV)   #this makes [16-255] to [16-235] at y plane
+YC = core.std.Expr(src16, expr=["x 219 255 / * 4096 +",""], format=vs.YUV)   #this makes [0-235] to [16-235] at y plane
```
Lypheo: This is functionally equivalent to the first one. Shouldn’t it rather be
```py
YC = core.std.Expr(src16, expr=["x 219 235 / * 4096 +",""], format=vs.YUV)
```
?

pwinner: Nope, it should be "x / 235 * 219 + 4096" or "x 235 / 219 * 4096 +" in reverse polish notation.

I think I made a mistake with the expression.

MysteryDove: Thanks! Fixed

encoding/video-artifacts.md
MysteryDove: I've added some explanation about the image.

MysteryDove:
Is it OK to merge this pull request?

FichteFoll:
I'll go over the text some time eventually and then merge. I'm not ready to merge as-is. Same for your other PR, btw, which I haven't looked into yet.

encoding/video-artifacts.md
Lypheo: Actually, wouldn’t it make more sense to use the built-in [``std.Levels``](http://www.vapoursynth.com/doc/functions/levels.html) filter instead of the whole Expr business?

FichteFoll: Wow, that looks really useful. I wonder why it doesn't accept lists for the `*_in` and `*_out` parameters that are expanded to the number of planes to be processed, like many other functions. Although the ones I'm thinking of probably accept multiple clips.

FichteFoll:
Please re-review, @Lypheo. I rewrote half of the content again, added more explanations where they were previously lacking, switched to `std.Levels` and tried to make sense of `overflow_notice.jpg`.

encoding/video-artifacts.md
```diff
@@ -232,6 +234,75 @@ or a manual filtering of dehalo\_alpha with dhhmask
...
+(and our filtering in general)
+with higher precision, 
+such as 16 bit or float (32 bit).
+For these clips,
+you would use the following[^5]:
+
+```py
+# 16 bit
+clip = clip.std.Levels(min_in=0, max_in=255 * 2**8, min_out=16 * 2**8, max_out=235 * 2**8, planes=0)      # y plane
+clip = clip.std.Levels(min_in=0, max_in=255 * 2**8, min_out=16 * 2**8, max_out=240 * 2**8, planes=[1,2])  # u&v planes
```
Lypheo: Bitshifting doesn't work for full range, since 255 << 8 = 65280 ≠ 65535. `max_in=2**16 - 1` would be correct.

encoding/video-artifacts.md
```diff
@@ -240,4 +311,8 @@ or a manual filtering of dehalo\_alpha with dhhmask
 
 [^3]: Blocking may also occur for other reasons other than compression data loss. [Image re-construction with padding][waifu2x238] can cause very similar looking effects, although this is irrelevant for fansubbing source videos.
 
+[^4]: The 8-bit limited range (used in rec.601 and rec.709) (and also BT.2020/2100) only defines values within \[16,235\] for the Y and \[16,240\] for the U and V planes. This means that Y=16 is considered full black and Y=235 full white, while any values outside of that range are clamped. U and V behave analogously.
```
Lypheo: Very minor nitpick: The standards permit out of range values up to a certain point, so they don’t necessarily need to be clamped. This is actually a good thing because it allows for a more accurate reconstruction of the original. For example, upscaling with kernels that have negative lobes (like Lanczos and some Bicubic variants) inherently introduces over-/undershoot; clamped values would thus adversely affect descale accuracy.

Edit: As an aside, I went ahead and tested my hypothesis from above, and you can actually verify it pretty easily:

```py
clip = core.std.StackHorizontal([core.std.BlankClip(width=25, height=150, color=[4096 if i % 2 else 235 * 2**8, 32768, 32768], format=vs.YUV444P16) for i in range(30)])
upscale = core.resize.Lanczos(clip, 1000, 150)
#upscale = core.std.Expr(upscale, "x 4096 max 60160 min")
out = descale.Descale(upscale, 750, 150, "Lanczos", taps=3)
```
You can clearly see the artifacts introduced by clamping (if you uncomment the third line).

encoding/video-artifacts.md
FichteFoll: Right, nice find. Also good point about bit shifting. Because I use that operator so rarely, I forgot about it.

encoding/video-artifacts.md
FichteFoll: Oh, I didn't mean the values where clamped when en-/decoding but during rendering. I should probably clarify on that.

PR #38, opened 2019-09-18
[On working][Help needed]The codecs page

MysteryDove:
I'm trying to explain how a codec work and combined them with the options in x264.

If there's any mistake or better way to explain, don't hesitate to make a commit.


FichteFoll:
FYI, I'll probably merge this onto a separate branch and do some hacking on it myself until I'm satisfied with the presentation. It'll take me some more time, so I probably won't be able to work in it until November. You are free to keep pushing commits, however.

PR #41, opened 2019-10-05
Write basics of x264 guide

shssoichiro:


shssoichiro:
I just saw that #38 exists, but fortunately it looks like there's not much overlap at the moment, as mine is a very intro-level guide. Going through each of the x264 options is what I intended to do for the Advanced section. @pwinner if you don't mind, I'd like to use your branch as a starting point for the Advanced section, but wanted to communicate first so there's no duplicate work going on (if you're still actively working on it).

MysteryDove:
Great! I'm considering the easy mode too. I'll keep pushing forward my part.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+x264 is a mature, free, open-source encoder for the H.264 codec.
+This guide will focus on x264.
+
+## Prerequisites
+
+To get started, you'll need two things:
+
+- A video to encode--for the examples,
+we will pipe in a video from VapourSynth,
```
FichteFoll: Please indent subsequent lines of a list.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
+# Encoding with x264
+
+H.264 has been the de facto standard video codec
```
FichteFoll: In line with the explanation in #38, we should differentiate between a codec and a video coding format. H.264 is the format, x264 the codec.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+TODO
+
+### Linux/macOS
+
+Generally, x264 will be available
+through your distribution's package manager.
+Here are a few examples:
+
+#### Ubuntu/Debian
```
FichteFoll: Formatting-wise, it's likely better to keep this concise in a simple list, e.g.:

```
- **Ubuntu/Debian**: `sudo apt-get install x264`
```

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+`sudo apt-get install x264`
+
+#### Arch Linux
+
+`sudo pacman -S x264`
+
+#### macOS
```
FichteFoll: The common classification has Windows, Linux and macOS (and BSD) on the same level, so this should be a h3.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+## Getting Started
+
+x264 is very configurable, and the options may seem overwhelming.
+But you can get started encoding by using the presets x264 provides
+and understanding a few basic concepts.
+We'll walk through those concepts with the following examples.
+
+## Example 1
+
+Go ahead and open up a terminal window,
```
FichteFoll: This is a comma splice.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+and navigate to the folder where your VapourSynth script lives.
+Let's run the following command:
+
+`vspipe --y4m myvideo.vpy - | x264 --demuxer y4m --preset veryfast --tune animation --crf 24 -o x264output.mkv -`
```
FichteFoll: Please use code blogs (with triple fences) when you're not using it inline.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+Let's run through what each of these options means:
```
FichteFoll: That's a good idea to go about it.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### `vspipe --y4m myvideo.vpy -`
```
FichteFoll: … but I'm not sure headings are the correct way to style this. Will probably have to see how this ends up looking to make a final call.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+with x264, our encoder.
+
+#### `--demuxer y4m`
+
+This tells x264 that we're providing it with a y4m file.
+This matches up with the `--y4m` flag that we gave to the `vspipe` command.
+
+#### `--preset veryfast`
+
+x264 has a set of presets to switch between faster encoding, or higher quality.
```
FichteFoll:
```suggestion
x264 has a set of presets to switch between faster encoding or higher quality.
```

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+You will almost never want to use the very extreme settings,
+but generally, if you want good quality
+and don't care about how long the encode takes,
+`slower` or `veryslow` are recommended.
+In this example, because we are just demonstrating how x264 works,
+we want a fast encode and have chosen `veryfast`.
+
+For the curious, you can see a full list of the settings
+enabled by each preset by running `x264 --fullhelp | less`.
```
FichteFoll: Except on Windows.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+in the current directory.
+
+The last argument to x264 is always the input file.
+In this case, we pass `-` for the input file,
+which tells x264 to use the piped output from vspipe.
+
+## Example 2
+
+For the next example, let's say we want to make sure our encode
+fits onto a single 4.7GB DVD. How would we do that in x264?
```
FichteFoll: Please split the lines here (and in the next paragraph).

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+This last portion tells which files to use for the input and output.
+We use `-o` to tell which filename to write the encoded file to.
+In this case, x264 will write a file at `x264output.mkv`
+in the current directory.
+
+The last argument to x264 is always the input file.
+In this case, we pass `-` for the input file,
+which tells x264 to use the piped output from vspipe.
+
+## Example 2
```
FichteFoll: Having a short title for what the examples are about helps finding the correct one later and gives a rough summary about their purpose.

E.g. `Example 2: Fixed File Size`

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+- The length of our video, in seconds.
+For this example, let's say our movie is 2 hours (120 minutes) long.
+We'll convert that to seconds: 120 \* 60 = **7200 seconds**.
+- Our target filesize.
+We know that this is 4.7GB,
+but we need to convert it to kilobits.
+We can do this with the following steps:
+  - 4.7GB \* 1024 = 4812.8MB
+  - 4812.8MB \* 1024 = 4928307.2KB (uppercase B = bytes)
+  - 4928307.2KB \* 8 = **39426457.6Kb** (lowercase b = bits)
```
FichteFoll: These equasions don't match the units on both sides. While some people think it's simpler to understand without the seemingly superfluous units, I prefer the explicit and correct sytle of `1024MB/GB` here, because it's self-documenting.

Also, while we're at it, these should be `GiB` and `MiB`.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+This tells x264 that we're providing it with a y4m file.
+This matches up with the `--y4m` flag that we gave to the `vspipe` command.
+
+#### `--preset veryfast`
+
+x264 has a set of presets to switch between faster encoding, or higher quality.
+The full list of presets, from fastest to slowest, is:
+
+- ultrafast
```
FichteFoll: Since this is an ordered list, we should also use an ordered list to better represent that. You can use `1.` for every line without counting up as the markdown converter will do that automatically.

FichteFoll (changes requested):
Thanks for the addition. This does has considerable overlap with #38, but I'm sure we'll be able to work it out. Fortunately there are no conflicts because it's in a different file.

I left some notes regarding formatting and correctness. The formatting notes with general advice (like code blocks) of course also apply to other locations.

FichteFoll:
How general are the concepts of 1-pass, 2-pass and CRF encoding? If they apply to many codecs (and I think they do), we might better move this to a the general encoding page and explain these concepts **with the example of x264**.

This page generally reads more like a guide than a reference, which I definitely like, but this might not be the correct page for it. You don't have to address this point within this PR, as I can work on how I imagine it later, but I'd like to talk about it first and hear some input.

MysteryDove:
> How general are the concepts of 1-pass, 2-pass and CRF encoding? If they apply to many codecs (and I think they do), we might better move this to a the general encoding page and explain these concepts **with the example of x264**.

There's QP, CQP, CRF, VBR, CBR in these codecs. I think they're a better start point than 1-pass, 2-pass, etc. Shall we introduce them first?

shssoichiro:
I agree, I think it makes sense to have an overview of the general concepts in the top level Codecs page, and guides/implementation details on the specific codec page (e.g. this one). So there's definitely some room to split up the 1-pass, 2-pass, and CRF sections into the main page.

encoding/codecs/h264.md
kageru: You could make it “Go ahead, open up a terminal window, and navigate […]”

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+we will pipe in a video from VapourSynth,
+which you should be able to do if you've been following
+the previous sections of this guide
+- The x264 encoder
+
+Here's how we get a copy of the x264 encoder:
+
+### Windows
+
+TODO
```
kageru: videolan serves the official builds here: https://download.videolan.org/pub/x264/binaries/win64/

encoding/codecs/h264.md
kageru: Perhaps a nitpick, but does it make sense to still use `apt-get` in guides or tutorials? `apt` was released years ago and has been in Ubuntu and Debian for at least 2 LTS versions.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### `vspipe --y4m myvideo.vpy -`
+
+This portion loads your VapourSynth script and pipes it to stdout,
+adding y4m headers that x264 can decode.
+If you use Linux, you're probably familiar with how piping works.
+If you're not, it's basically a way of chaining two commands together.
+In this case, we want to chain `vspipe`,
+the program that reads VapourSynth scripts,
+with x264, our encoder.
```
kageru: but `x264` in code blocks to be consistent with `vspipe` 2 lines earlier

encoding/codecs/h264.md
kageru: +1

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+- veryfast
+- faster
+- fast
+- medium
+- slow
+- slower
+- veryslow
+- placebo
+
+You will almost never want to use the very extreme settings,
```
kageru: “very extreme” seems redundant

encoding/codecs/h264.md
kageru: as a sidenote: `--fullhelp` is outdated for many things. The presets should be fine, though.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+It is generally not recommended to go higher than CRF 30 in any real-world
+encoding scenario, unless you want your videos to look like they were made
+for dial-up.
+
+#### `-o x264output.mkv -`
+
+This last portion tells which files to use for the input and output.
+We use `-o` to tell which filename to write the encoded file to.
+In this case, x264 will write a file at `x264output.mkv`
+in the current directory.
```
kageru: Maybe mention the significance of the extension here. x264 can generate raw bytestreams when the output ends with `.264` or mkv/mp4 if it was compiled with the respective flags.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+The last argument to x264 is always the input file.
+In this case, we pass `-` for the input file,
+which tells x264 to use the piped output from vspipe.
```
kageru: It’s not the last argument; it’s the only position argument. You could put it anywhere.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+
+For the next example, let's say we want to make sure our encode
+fits onto a single 4.7GB DVD. How would we do that in x264?
+
+First, we'll need to figure out what bitrate our encode should be,
+in **kilobits per second**. This means we'll need to know a couple of things:
+
+- The length of our video, in seconds.
+For this example, let's say our movie is 2 hours (120 minutes) long.
+We'll convert that to seconds: 120 \* 60 = **7200 seconds**.
```
kageru: I’d prefer a codeblock for the calculation.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+- CRF 16-18: This is considered visually lossless to most viewers,
+and leans toward high quality while still providing a reasonable filesize.
+This is a typical range for fansub encodes.
+- CRF 21-24: This provides a good balance between quality and filesize.
+Some quality loss becomes visible, but this is generally a good choice
+where filesize becomes a concern, such as for videos viewed over the internet.
+- CRF 26-30: This prioritizes filesize, and quality loss becomes more obvious.
+It is generally not recommended to go higher than CRF 30 in any real-world
+encoding scenario, unless you want your videos to look like they were made
+for dial-up.
```
kageru: What feels odd about this section is that the ranges aren’t continuous. What about the range between 13 and 16? 19 and 20? Bluray encodes often are in the range 13-15, so you could just extend these ranges a little to not leave any gaps.

encoding/codecs/h264.md
```diff
@@ -0,0 +1,247 @@
...
+but it generally helps to produce a better-looking video.
+
+#### `--crf 24`
+
+CRF is a constant-quality, 1-pass encoding mode.
+In layman's terms,
+this means that we don't need the output to meet a specific filesize,
+we just want the output to meet a certain quality level.
+CRF ranges from 0 to 51,
+with 0 being the best quality and 51 being the smallest filesize,
```
kageru: it’s actually -12 to 51. -12 is the equivalent of using `--qp 0`.

encoding/codecs/h264.md
kageru: And I much prefer `Kbit` over `Kb`. That would probably make the explanation of uppercase/lowercase unnecessary as well.

encoding/codecs/h264.md
FichteFoll: I considered this too, but decided the reader should be smart enough to recognize that anything outside the regions specified is either, well, not in any of these regions or only for special occasions. It's hard enough to define definite regions for CRF as is already, so leaving a few gaps leaves more room for interpretation.

Now, what we should mention is that CRF is slightly dependant on resolution, or rather what videophiles generally consider as transparent or acceptable.

encoding/codecs/h264.md
Lypheo: Only for 10 bit. What he said is true for 8 bit.

encoding/codecs/h264.md
shssoichiro: 90% of the guides I see still use `apt-get`. But guides like [this](https://itsfoss.com/apt-vs-apt-get-difference/) make me think that's outdated and `apt` is better. Feels like searching for answers on StackOverflow and the first one you find is from 2012.

encoding/codecs/x264.md
```diff
@@ -25,7 +25,12 @@ Here's how we get a copy of the x264 encoder:
 
 ### Windows
 
-TODO
+Official Windows builds are available
+[here](https://download.videolan.org/pub/x264/binaries/win64/).
+For these examples,
+you'll want the latest version
+of the standard 8-bit build,
+which is the version without "10b" (for 10-bit) in the name.
```
kageru: The binaries were unified some time ago. That’s the reason there are no 10-bit binaries after r2851. Depth is now controlled by `--output-depth`

encoding/codecs/h264.md
kageru: My bad, then. Another case of `--fullhelp` not telling the truth.

encoding/codecs/h264.md
OrangeChannel: image
image

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+4812.8MiB * 1024KiB/MiB = 4928307.2KiB
+4928307.2KiB * 8Kbits/KiB = 39426457.6Kbits
+```
+
+Now we divide the kilobit size we calculated by our video length,
+to find our kilobit per second target bitrate:
+
+```
+39426457.6Kbits / 7200 seconds = 5475 Kbps
+```
```
OrangeChannel, line 228: We have the option of changing this math to render with TeX (which also might come in handy on other pages where math has been put in codeblocks or in-line currently) with the `katex` gitbook plug-in. Here's what the changes would look like (and the rendering works with the current text/theme options we give the user).

### Currently:
image

### With katex:
image
image

I know @FichteFoll has mentioned concerns about adding more plug-ins to the book, so I'll ask for their feedback on this.

---

With the current 12 plug-ins, initial generation: `finished with success in 2.0s`
"" on file-changed restart: `finished with success in 1.7s`

With katex, initial generation: `finished with success in 2.5s`
"" on file-changed restart: `finished with success in 1.9s`

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+H.264 has been the de facto standard video format
+across the internet for the past decade.
+It is widely supported for playback in all modern browsers
+and many hardware devices such as gaming consoles and phones.
+It provides better video quality at smaller file sizes
+compared to its predecessors.
+
+x264 is a mature, free, open-source encoder
+for the H.264 video format.
+This guide will focus on x264.
```
OrangeChannel: I think this line is redundant, or should at least be changed to say `This page`

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Prerequisites
+
+To get started, you'll need two things:
+
+- A video to encode--for the examples,
```
OrangeChannel: An em-dash instead? —

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+and don't care about how long the encode takes,
+`slower` or `veryslow` are recommended.
+In this example,
+because we are just demonstrating how x264 works,
+we want a fast encode and have chosen `veryfast`.
+
+For the curious,
+you can see a full list of the settings enabled by each preset
+by running `x264 --fullhelp | less` (Linux/Mac)
+or `x264 --fullhelp | more` (Windows).
```
OrangeChannel, line 118: We should probably mention that fullhelp isn't fully up-to-date (and it likely won't be updated) in a footnote here.

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+
+x264 is very configurable,
+and the options may seem overwhelming.
+But you can get started encoding
+by using the presets x264 provides
+and understanding a few basic concepts.
+We'll walk through those concepts
+with the following examples.
+
+## Example 1: General-Purpose Encoding
```
OrangeChannel: These examples should be on a third-level heading I think, as `Getting Started` is 2 and the commands are 4.

### Currently
image


### Changed: 
image

### what it would look like if the commands were changed to level 5
image

OrangeChannel (changes requested):

encoding/codecs/x264.md
```diff
@@ -0,0 +1,292 @@
...
+1. ultrafast
+1. superfast
+1. veryfast
+1. faster
+1. fast
+1. medium
+1. slow
+1. slower
+1. veryslow
+1. placebo
```
OrangeChannel, line 105: This numbering is *actually* okay because it wraps this in an `<ol>` HTML tag, so the numbers here don't matter.

shssoichiro: Yeah, this is the standard way of doing ordered lists in markdown. It makes it easier to add/remove items in the future (though that probably won't happen in this case).

OrangeChannel: I should probably update this on the contributing page, since this is much easier and cleaner looking :+1: .

FichteFoll: Good idea. I didn't check whether we mention this already, but I use this style in a couple other documentation projects. (Just like semantic line feeds, and I think everyone should use those!)

encoding/codecs/x264.md
FichteFoll: I love LateX. Let's do this! Thanks for taking the time to look into (as I considered suggesting this as well but due to being relatively short on time I didn't want to commit to it).

encoding/codecs/x264.md
FichteFoll: :+1:, also in at least one other location.

FichteFoll (approved):

FichteFoll:
This PR is a prime example of how I imagine this collaborative effort to work. :+1: Thanks for the help on this, everyone, and of course @shssoichiro. I'm satisfied with this as is and am going to merge now.

---

@OrangeChannel, please do the katex integration in a separate PR. You can add a sentence or two regarding ordered lists on the master branch directly.

PR #42, opened 2019-10-10
Add gitbook plug-in "katex"

OrangeChannel:


OrangeChannel:
e1ea810d3b8b153accd46a6f1e927a8bad3e360a has some weird stuff so I'll explain it here. On GitHub, it doesn't change the `$$` within back-ticks to the templates, but gitbook *always* does. That explains the weirdness of me explaining the templates, but there not being any templates in the actual markdown.
There's no actual way around this, as you can't escape characters within back-ticks.

encoding/masking-limiting-etc.md
```diff
@@ -400,12 +401,12 @@ integer and float formats,
 so for more complex filtering
 float is recommended whenever possible.
 In 8 bit integer format where neutral luminance (gray) is 128,
-the function is `clip1 - clip2 + 128` for MakeDiff
-and `clip1 + clip2 - 128` for MergeDiff,
+the function is $$\big( \mathrm{clip\, a} - \mathrm{clip\, b} + 128 \big)$$ for MakeDiff
```
FichteFoll: Why do you use `\big` here?

encoding/resampling.md
```diff
@@ -624,6 +624,6 @@ shifted_scaled_u = core.resize.Spline16(u, 1920, 1080, src_left=0.25) # shifts t
...
+[^4]: This is derived as follows: The shift is the distance between the position of the first luma sample and the position of the first chroma sample (both mapped onto the input grid and given in terms of input chroma pixel widths). The former is located at $$\left( 0.25 + \frac{\mathrm{src~width}}{4 \times \mathrm{dst~width}} \right)$$, the latter at $$\left( \frac{\mathrm{src~width}}{2 \times \mathrm{dst~width}} \right)$$. This yields $$\Big[ 0.25 + \frac{\mathrm{src~width}}{4 \times \mathrm{dst~width}} - \frac{\mathrm{src~width}}{2 \times \mathrm{dst~width}} = 0.25 + \frac{\mathrm{src~width}}{\mathrm{dst~width}} \times (^1/_4 -\,^1/_2) = 0.25 + \frac{\mathrm{src~width}}{\mathrm{dst~width}} \times (-0.25) \Big]$$ for the shift.
```
FichteFoll: Again, why `\Big`? Also sure would be nice if footnotes could have line breaks. 😞

FichteFoll:
You could also format the footnote I added regarding limited 8-bit range in artifacts.md

encoding/masking-limiting-etc.md
OrangeChannel: #### big
image
#### normal parens
image

It's subtle but I think it helps to separate the formulas from the text better. Originally, the in-line would've had a dark background to separate it.

#### original
image

FichteFoll: Well, that should have read "Why do you use `\big(`?", but I see that it's probably to make the formula stand out more from the text.
Can we target them with a CSS selector instead?

encoding/resampling.md
OrangeChannel: Similar to the other response, it helps separate it (also good for line height when it's placed below due to sizing changes) from the text.
#### Big
image
image

#### normal `[ ]`
image
image

#### old
image
image

OrangeChannel: Could maybe even use an non-breaking space inside on both sides

encoding/masking-limiting-etc.md
OrangeChannel: Yea, we actually can. What did you have in mind?

OrangeChannel: I could add a background to in-lines like they originally had

encoding/resampling.md
FichteFoll: Hm, with a long formula like this, we should probably try looking into using an info box instead of a footnote. It's still extra information that's not necessary to grasp the content, but a footnote probably isn't the best place for ti.

encoding/masking-limiting-etc.md
FichteFoll: Yeah, I think something for the background would be nice. A semi-transparent black/white, depending on the theme. 

Doesn't *need* to be the same as for inline code blocks, but I guess choosing those would at least be consistent.

encoding/video-artifacts.md
```diff
@@ -311,8 +311,8 @@ simply adjust the `min/max_in/out` parameters accordingly.
...
+[^4]: The 8-bit limited range (used in rec.601 and rec.709) (and also BT.2020/2100) only defines values within \[16,235\] for the Y and \[16,240\] for the U and V planes. This means that Y=16 is considered full black and Y=235 full white, while any values outside of that range are clamped virtually (during rendering). U and V behave analogously.
 
-[^5]: The limited ranges in different precisions are shifted by (multiplied by 2 to the power of) the added bits. For 12-bit, for example, you muliply by 2\^(12-8), resulting in \[256,3760\] and \[256-3840\] respectively. The maximum value in full range is obviously the highest unsigned integer value, so 2\*\*12-1.
+[^5]: The limited ranges in different precisions are shifted by (multiplied by 2 to the power of) the added bits. For 12-bit, for example, you multiply by $$2^{12-8}$$, resulting in \[256,3760\] and \[256-3840\] respectively. The maximum value in full range is obviously the highest unsigned integer value, so $$2^{12-1}$$.
```
FichteFoll: `$$2^{12}-1$$`. Also, I think you could katexify the ranges, too.

FichteFoll (changes requested):

encoding/video-artifacts.md
OrangeChannel: image
there's the three style options for the ranges, what we have now, sans-serif, and normal tex math font. I think I like what we have now better actually.

OrangeChannel:
Playing around with CSS right now, we have the option of using the current font on the page instead of the TeX font.

tex:
image

website serif:
image
website sans:
image

Let me know what you think @FichteFoll .

encoding/video-artifacts.md
FichteFoll: I prefer the sans-serif version, because this isn't a print medium.

FichteFoll:
Using the page's font seems nice as it will be less of a context break than a separate paragraph already is. And using sans-serif on a website should always be preferred (unless the user dislikes that for some reason).

FichteFoll:
How does this work with katex? Does it generate svg?

OrangeChannel:
So, I chose to go with the website font theming for generic math font for KaTeX stuff.
image
It changes to serif if you use the dropdown selector at the top.
The block code/TeX also share the same theming now, except for a slightly larger non-mono font.
image

OrangeChannel:
> How does this work with katex? Does it generate svg?

No, in fact the benefit of KaTeX over other *tex libraries is that it pre-renders them as HTML at build-time, leaving no shape/svg generation on the user side.

https://katex.org/

In the case of changing font-styles, that’s all it’s actually doing. The dropdown is just a css change, no html/svg differences

FichteFoll (approved):

FichteFoll:
Thanks for taking the time to work on this. You're the only person who has touched our CSS so far, so I'll trust your judgement when you say it's easier to edit like this.

PR #43, opened 2019-10-10
CSS: small fix for portrait screens

OrangeChannel:
- also includes a fix for a typo in the image scaling section

### DIFF
--- a/styles/website.css
+++ b/styles/website.css
@@ -453,6 +453,13 @@ li.header:nth-child(1) {
         padding: 1.5em 1.5em 3em 1.5em;
     }
 }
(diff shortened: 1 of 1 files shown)


styles/website.css
```diff
@@ -539,7 +546,7 @@ img {
 }
 @media (max-width: 1240px) {
     .markdown-section img {
-        max-width: 100%
+        max-width: 100%;
```
FichteFoll, line 549: Trailing semi-colons are actually optional (maybe not by the spec but probably with all relevant implementations), but this doesn't hurt.

FichteFoll (approved):

PR #47, opened 2019-10-18
Replace ADC with working tracker

cshmnyfy:
I don't use the tracker myself but since ADC is dead for quite some time now, it might be a good idea to add an alternative.

### DIFF
--- a/encoding/preparation.md
+++ b/encoding/preparation.md
@@ -29,10 +29,10 @@ If you don't have a reliable way to get raws
 and if your group doesn't provide them,
 try finding a source first.
 Private bittorrent trackers like [u2][]
-or [AsianDVDClub][] are good starting points.
+or [SkyeySnow][] are good starting points.
 
 [u2]: https://u2.dmhy.org "u2"
-[AsianDVDClub]: https://asiandvdclub.org "ADC"
+[SkyeySnow]: https://skyeysnow.com "SkyeySnow"
 
 
 ## Processing and Filtering


FichteFoll:
I'm not in the raw business, so I'd like to hear someone else's opinion on which tracker is a good altrenative to u2. cc @LightArrowsEXE

LightArrowsEXE:
For public trackers, Nyaa's all you need. The only good private tracker I'm aware of is u2, although people seem to like AnimeBytes as well. Skyeysnow is also an up- and coming public tracker that people seem to be using as of late, with some stuff even popping up there first (like Case Files vol.1). Any of those will work as an alternative now that ADC is down.

EDIT: actually saw the commit now, it should be noted that, as far as I could tell when I signed up, skyeysnow is a _public_ tracker. Not a private one.

FichteFoll:
>Welcome to SkyeySnow:
>
>「天雪」 is a forum built by the desire of some of the release groups to have a good sharing environment.
>
>In terms of sharing，Skyey use Private Tracker(PT) and Net-Disk
>
><About PT>
>
>A full-featured PT is embedded in Skyey as the main part of sharing
>PT downloads will consume points, If the seeding rate (related to the average seeding time) or the sharing rate (related to the upload amount) is low, there will be some  punishment  in download preferential aspects.As the low seeding rate punishment is more serious. If your network is not that good. You should have a the average seeding time reach 100 hours.
>There are some help here https://skyey...com/forum.php?mod=forumdisplay&fid=38
>
><About Net-Disk>
>Skyey also has a network disk forum to share resources, but only have Chinese rules now.
>
><About subtitles>
>Sharing English Subtitles in skyey please use google drive share in the English Forum
>See the top of here https://skyey...com/forum.php?mod=forumdisplay&fid=38
>
>
>END:
>
>Once again,  Welcome to SkyeySnow
>Wish you all have a happy time here  (╯°Д°)╯︵ ┻━┻ 

This was in the PN I had in my inbox. Kinda hard to decipher whether it's public or private, but doesn't look too public to me.

cshmnyfy:
According to their own site, SkyeySnow is indeed a private tracker. They just (currently) happen to have open signups. Also since this guide aims to become a public guide for aspiring fansubbers, I think it would be nice to list a tracker that is actually accessible.

PR #49, opened 2019-10-26
[WIP] [RFC] Mastroka Tricks Page

OrangeChannel:
This is **not** finished yet, want to get some feedback on how to write this, as I don't see a way of doing this without a step-by-step tutorial (which idk, may not fit the guide). Let me know what shouldn't be included if it already is, or what should be included besides the stuff I have listed at the top.

It's been massively cleaned up with the octrim() function, so this shouldn't become as long as it originally was.

Also unsure of where it even fits on the sidebar.


encoding/mkv.md
```diff
@@ -0,0 +1,464 @@
...
+- `00004.m2ts` > `00004.wav`
+- `00005.m2ts` > `00005.wav`
+- `00007.m2ts` > `00007.wav`
+- `00008.m2ts` > `00008.wav`
+
+Now, either in a new python instance
+or from within our VapourSynth script above,
+we will use `acsuite.py` to trim our audio,
+and generate chapter timings.
+(The lines should be run one at a time,
```
OrangeChannel: I actually fixed this, forgot to push the changes to acsuite though

OrangeChannel:
Re-opened with a better re-write in <https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/54>.

PR #51, opened 2019-10-29
Fix bitrate calculation

OrangeChannel:
If we were to use GiB instead, the answer still wouldn't have been 5475 kbps as 1Kib != 1Kb: 1Kib is 1.024Kb, so the answer should've been ~5607 kbps.
```py
>>> find_filesize(5475, seconds=7200)
estimated filesize is 4.59 GiB or 4.93 GB
>>> find_filesize(5607, seconds=7200)
estimated filesize is 4.70 GiB or 5.05 GB
```
I checked that the DVD size is specified in decimal GB and listed the source.

Added the python script (that I initially wrote for the ordered-chapters math) so nobody has to do this manually anymore.

### DIFF
--- a/encoding/codecs/x264.md
+++ b/encoding/codecs/x264.md
@@ -196,7 +196,7 @@ as the only argument without a `--` flag before it.
 
 For the next example,
 let's say we want to make sure our encode
-fits onto a single 4.7GB DVD.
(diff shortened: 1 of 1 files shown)


shssoichiro:
Whoops, guess I probably should've checked that. :laughing:

PR #54, opened 2020-02-15
[WIP][RFC] Ordered-chapters walkthrough/guide

OrangeChannel:
This is still unfinished, but figured might as well post this here for more visibility for people not in the Discord.

TODO:

- [ ] completely re-write the whole thing with the libbluray advice from thebombzen :disappointed: 

- [ ]  fix up the `TODO`s in the md
- [ ] explain editions
- [ ] use editions to swap between OP/ED and NCOP/NCED
- [ ] explain how to create a playall file

TODO in actual script:

- [x] Allow handling of both audio and video while trimming/splicing
- [x] Write chapter timestamps to valid XML
- [ ] Find a better way to support VFR input clips (probably should require timecodes file with fallback on acsuite's implementation)
- [ ] Re-write tests to cover all new functionality
- [ ] Check that multiple OP/ED work as expected (should but just incase because this isn't in the current tests)

Blockers:
- [ ] mpv # 7963 about playall files not loading attachments properly
- [ ] vapoursynth audio support branch for new frames method for VFR clips and possibly some audio stuff


encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+The Mastroka Media Container is an open source,
+open standard multimedia container
+that is heavily used in the fansubbing scene.
+Mastroka video, or `.mkv` files
+are the standard for soft-subbed releases
+as they allow for multiple subtitles tracks and chapters
+which are a somewhat unique feature to the container.
+The MPEG-4 Part 14 or MP4 is another container format
+but it is limited in its usefulness for fansubbing.
```
FichteFoll, line 17: Imo we have enough *time* to explain why that is the case, unless this is already mentioned somewhere else in the guide and we can just refer to it.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+The following steps will require
+VapourSynth,
```
FichteFoll: Why is vapoursynth not linked?

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+will also be required for testing the files.
+
+The examples will be using pre-made subtitle tracks,
+a BDMV source,
+[VSEdit](preparation.md#the-editor),
+an [audio-cutting Python script][acsuite.py],
+and a multitude of utility scripts/functions to help speed up the process.
+
+What we will be doing:
+1. Extracting audio tracks from BDMV .m2ts files
```
FichteFoll: Newline between these.

Also, it seems more natural to me to write 
```
We will: 

1. Extract …, 
1. Trim, 
… , and
1. Create …
```
but YMMV.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+[toolnix]: https://mkvtoolnix.download/downloads.html
+[FFmpeg]: https://ffmpeg.zeranoe.com/builds/
+[mk-playback]: https://github.com/hubblec4/Matroska-Playback/blob/e48149b500ce80c38ebc07ccf8f74f8af45e89b2/src/PlayerOverview.md
+[MPC-HC]: https://github.com/clsid2/mpc-hc/releases
+[mpv]: https://mpv.io/installation/
+[acsuite.py]: https://github.com/OrangeChannel/acsuite
+
+
+### Starting with a clean work environment
```
FichteFoll:
```suggestion
### Preparing the work environment
```

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+The following steps should be done from within a folder containing
+all of the files (episodes) re-named to some simple format like `ep##.m2ts`,
+as having multiple `00004.m2ts` files won't be helpful.
+The example scripts provided will assume this workflow.
+
+
+## Extracting the audio
+
+If the source files you are working with are not .m2ts,
```
FichteFoll:
```suggestion
If the source files you are working with are not `.m2ts`,
```

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+re-encoding your audio into PCM / WAV is *still* recommended
+as slicing some lossless or lossy codecs is not precise.
```
FichteFoll: What about FLAC here? Are there lossless codecs that don't decode or losslessly convert to WAV/PCM? All others should be easily spliceable.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+ffmpeg -i ep01.m2ts -map a:0 ep01.wav
+```
+
+If the desired audio track is not the first,
+`-map a:#` should be changed to the track # shown by `ffprobe`.
+
+*If you are following along with the next few steps,
+you don't need to extract the audio just yet.
+The audio-cutting script will apply this FFmpeg command to the files
+automatically.*
```
FichteFoll: This (and later similar occurances of full-paragraph italics) would be a good candidate to appear in a note box. I don't know/remember if gitbook has something like that by default, but there must at least be a plugin to do this and I'd like to have the possibility of using those throughout the guide.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+[tsMuxeR]: https://www.videohelp.com/software/tsMuxeR
+
+
+## Starting with trimming the video
```
FichteFoll: Beginning a section with "Starting" is weird. This should be rephrased.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+Starting with the BDMVs,
+our first step is to trim the main episodes out.
+This requires knowing the frame ranges, shown below.
+Second, we will crop out the NCOP/NCED files
+as they also have unwanted frames at the end.
+
+The first time the OP appears,
+we will also want to note this frame range
+ and on what episode it occurs,
```
FichteFoll: Leading space here.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+```py
+import vapoursynth as vs
+core = vs.core
+
+bdmv01 = core.lsmas.LWLibavSource('ep01.m2ts')
+bdmv02 = core.lsmas.LWLibavSource('ep02.m2ts')
+...
+bdmv13 = core.lsmas.LWLibavSource('ep13.m2ts')
+bdmvNCED = core.lsmas.LWLibavSource('nced.m2ts')
+bdmvNCOP = core.lsmas.LWLibavSource('ncop.m2ts')
```
FichteFoll: I personally find it pretty weird to do all of this in a single file instead of the normal per-episode encode script and it looks like you need to convert these trims into separate function calls for acsuite later anyway.

Regardless, the way I currently understand it it would lead to having to keep two scripts in sync regarding the trims you make and might adjust later.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+- `ep01.wav` (delete)
+- `ep02.wav` (delete)
+- `nced_cut.wav` *
+- `nced.wav` (delete)
+- `ncop_cut.wav` *
+- `ncop.wav` (delete)
+- `op_cut.wav` *
+
+Files marked with an asterisk * are our trimmed audio files.
+The rest are artifacts from the FFmpeg extraction that can be deleted now.
```
FichteFoll, line 263: Is the script aware of these intermediate files for caching purposes? If it is not, you might as well delete them from the script.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+For `octrim`, we will need the chapter frame ranges (inclusive)
+and optionally their names. The frames here are from the **untrimmed source
+clips (bdmv01, bdmv02, etc.)**.
+
+Putting these into a Python script as lists of 3-tuples (start, end, name)
+will help later when trimming the audio.
+
+*Note: for `chaps01`, `'Part A'` starts on frame 0 and ends on frame 3770.
+The ending number can be left off
```
FichteFoll, line 294: Explain the usual case first and the odd case later.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+```py
+import acsuite as ac
+ac.audio_trim('ep01.m2ts', chaps01)
+ac.audio_trim('ep02.m2ts', chaps02)
+...
+ac.audio_trim('ep11.m2ts', chaps11)
+ac.audio_trim('ep13.m2ts', chaps13)
+```
+
+Running the above script with Python (may take a few minutes to complete)
```
FichteFoll, line 330: Parentheses in new line.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+- `ep1_cut.wav` -- `ep13_cut.wav` (total: 2.9 GiB)
+- `(ed|op|nced|ncop)_cut.wav` (total: 65.6 MiB)
+
+[acshelp]: https://github.com/OrangeChannel/acsuite/blob/a7e36aa65533a93bbaaa75b6fc5345194b27a5df/acsuite.py#L127
+
+
+## Encoding the audio
+
+We will be using FFmpeg's *libopus* and *flac* encoders.
+The following examples simply loop[^2] the command for every `ep*.wav` file,
```
FichteFoll, line 344:
```suggestion
The following examples simply run[^2] the command for every `ep*.wav` file,
```

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+and all four op/ed files.
+
+For the main episodes, we will be using opus:
+
+```sh
+$ fd -e wav --regex "(_cut)" -E 'op*' -E 'ed*' -E 'nc*' -x ffmpeg -i {} -c:a libopus -b:a 128k -cutoff 20000 {.}.opus
```
FichteFoll, line 350: You don't do anything with the capture and `--regex` is the default, so just use `fd .e wav '_cut' …`. Besides, if you exclude `*op*` and `*ed*` you can skip the nc exclude.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+This results in the following files:
+
+- `ep1_cut.opus` -- `ep13_cut.opus` (total: 259.7 MiB) 91.3% reduction
+- `(ed|op|nced|ncop)_cut.flac` (total: 39.8 MiB) 39.3% reduction
+
+Normally, using two different audio codecs within the same file
+would not be possible.
+However, ordered chapters make it possible for each file to use its own codec
+(as long as the track order is the same).
```
FichteFoll, line 367: Anything about compatibility for this?

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+the filesize difference and lossless music should satisfy both groups.
+Normal dialogue should be nearly transparent at 128 kbps with opus,
+but this comes with some caveats.
+If there are 'insert' songs within an episode that you'd like to keep lossless,
+you will need to encode the entire episode with a lossless codec.*
+
+
+## Working with ordered chapters
+
+*At this point, you should have encoded clips of each episode
```
FichteFoll, line 384: This paragraph doesn't need to be in italics, imo.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+Before we can start using ordered chapters,
+we will have to mux the current files into mkv's.
+This is because ordered chapters reference other mkv's by their UID
+to generate the virtual timeline for the player.
+MKVToolNix can save chapters *into* existing mkv files,
+so this step will be the **only time** we mux.
+
+This can be done with the MKVToolNix GUI,
+in which case the order of the tracks in the bottom list is important.
+Although mpv will only warn that tracks are mismatched,
```
FichteFoll, line 411:
```suggestion
Although mpv only emits a warning when tracks are mismatched,
```

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+it is a good idea to keep the same track order in all of the mkv's.
```
FichteFoll, line 412: Pretty sure this is not how you pluralize abbreviations, but someone more knowledgeable should verify this.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+video_op = ['ed', 'nced', 'ncop', 'op']
+videos = video_eps + video_op
+
+audio_eps = ['ep{:02d}_cut.opus'.format(i) for i in range(1, 14)]
+audio_op = ['ed_cut.flac', 'nced_cut.flac', 'ncop_cut.flac', 'op_cut.flac']
+audios = audio_eps + audio_op
+
+for n, a in zip(videos, audios):
+    args = shlex.split('mkvmerge -o {0}.mkv {0}.264 {1} {0}.ass'.format(n, a))
+    subprocess.run(args)
```
FichteFoll, line 438: Same as discussed in discord. :wink: Build a list and run that. Also, the internal variable name for the first parameter is `cmd` and that better fits what we're building here.

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+Instead, we will simply use the number of frames in the entire op
+to determine the end timestamp.
+
+From our trim, `op = bdmv[30042:32201]`,
+we find that the clip is $$32201-30042=2159$$ frames long.
+This can also be found from `op.num_frames` in VapourSynth.
+
+Using a small [script to change frame numbers into timestamps][f2ts.py],
+we find that 2159 frames translates to `00:01:30.048291667`.
+This can also be accomplished with this following line in a Python instance.
```
FichteFoll, line 514: What does "Python instance" refer to exaclty?

encoding/mkv.md
```diff
@@ -0,0 +1,505 @@
...
+
+```py
+for _ in input().split():print('%02d:%02d:%09f'%((m:=(s:=int(_)/23.976)//60)//60,m%60,s%60))
```
FichteFoll, line 517: You should *never* use `_` as a variable name because that has special meaning in interactive sessions and is used as a "drop this" convention in scripts.

FichteFoll (changes requested):
Did an initial review that took me much more time than I hoped to spend on it.

I generally miss a short paragraph or at least a TL;DR on *why* one would even *want* to use Ordered Chapters. It is very important to provide context to decisions and not just explain how to apply them.

encoding/mkv.md
OrangeChannel: Yea, it is but how should I remove them cross-platform?

encoding/mkv.md
OrangeChannel: It seems to be 
```py
def run(args: Union[bytes, str, Sequence[Union[bytes, str, PathLike]]], ...
```
so maybe what you're thinking of is some other library or an older version of subprocess?

encoding/mkv.md
FichteFoll: I don't see what part of this argument is cross-platform? `os.unlink` works on all. Regardless, if these files can be considered a cache, you can leave them be or cache them in a subdirectory to keep things more organized. From a workflow standpoint, however, I hate having to delete intermediate files I'm not going to look at again and rather have the tool that created them do that for me.

encoding/mkv.md
FichteFoll: Maybe I was confusing things, then. I still find "args" to be a weird choice because it obviously contains the executable as the first element, but my argument regarding the internal name seems to be debunked.

OrangeChannel:
> Did an initial review that took me much more time than I hoped to spend on it.
> 
> I generally miss a short paragraph or at least a TL;DR on _why_ one would even _want_ to use Ordered Chapters. It is very important to provide context to decisions and not just explain how to apply them.

@FichteFoll 

---
Ordered-chapters are a special feature of mkv files that allows
the player to reference other mkv files in the same folder
to create a virtual combined timeline.

The example use cases we will be looking at starts with the ability
to cut out the OP/ED from each episode in a series 
(assuming the OP/ED stays consistent throughout the show).
This comes with at least four benefits that are unachievable
using normal chapters:

1. We avoid having to encode the same OP/ED sequence for each episode
   saving us encoding time.
1. We can use different audio/video settings for the OP/EDs
   than for the main episodes.
    - Using FLAC for the music and a more compressible lossy codec
      for the main episode audio tracks will save space and satisfy audiophiles
      that want lossless music.
    - Using a much lower CRF value for the OP/ED will still likely save space
      as we are only encoding each once.
1. We can create a 'playall' file that creates a virtual timeline of
   all of the main episodes, skipping previews and
   only playing the OP/ED once over the entire timeline.
1. Using editions, we can allow the viewer to switch between the no-credit (NC)
   versions of the OP/ED and the normal ones.
---

You think this is too long?

encoding/mkv.md
OrangeChannel: I'm going to leave the testing up to someone else for the time being. The whole 'not starting with a BDMV source' part of this page is gonna be a TODO for some time.

OrangeChannel:
@FichteFoll Regarding noteboxes, I can't find a plugin so I made a new class with some basic CSS (the dashed border is because I can't decide on colors for each theme right now).
image
image
image
The last (white theme) pic shows 100% width forced. In theory, it would be slightly different color than the background and there'd only be the border on the left. It uses the font-awesome font that we seem to already be loading on the website for other things.

EDIT: here's a rough mobile design
image

OrangeChannel:
Just for reference:
```css
.infobox {
    display: inline-flex;
    border-left: .4em solid;
    border-top: .1em dashed;
    border-bottom: .1em dashed;
    border-bottom-left-radius: .2em;
    border-top-left-radius: .2em;
    border-bottom-right-radius: .2em;
    border-top-right-radius: .2em;
    border-color: inherit;
    opacity: 80%;
    border-right: .1em dashed;
    padding-left: 0;
    padding-top: .5em;
    padding-bottom: .5em;
    padding-right: .5em;
    margin-top: -.25em;
    margin-bottom: -.25em;
    background: 0 0;
    width: 100%
}

.infobox p {
    padding-top: .5em;
    font-size: .85em;
    font-style: italic;
    margin-bottom: 0;
    padding-bottom: .5em
}

.infobox::before {
    content: "";
    display: inline-flex;
    font: 14px/1 FontAwesome;
    font-size: 1.75em;
    margin: 0 .5em 0 .8em;
    align-self: center
}
```
```md
<div class="infobox"><p>
If you are following along with the next few steps,
you don't need to extract the audio just yet.
The audio-cutting script will apply this FFmpeg command to the files
automatically.
</p></div>
```

encoding/mkv.md
OrangeChannel: See my comment on the PR.

FichteFoll:
> You think this is too long?

No, I think that's great.

Info box also looks good, except for entire text being in italics and noticeably smaller font size. I would definitely prefer using the normal font size or at least larger than that and I'm not a fan of using italics like this because it makes the strokes thin and harder to read with no added benefit. We already visually mark the entire box, so we wouldn't need to do it for the text as well.

---

For the sake of reviweing the PR and general workflow, please don't force-push already reviewed commits, so I can focus on the changes after than and don't have to rely on Github providing [direct comparisons](https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/compare/fa1367ef077dc95161cb27e383581639a29f9f88..4edf71aaa6b2110f8d72c97869be7725c355c34f) with the lost head.

I'll review the changes later.

OrangeChannel:
You might've missed the conversation I had in Discord with thebombzen but pretty much this is getting re-re-written from scratch (it was in the checklist at the top of the PR before you reviewed) so a lot of the stuff you reviewed is probably going to become irrelevant as I continue working on this. But if it makes it cleaer I'll just commit on top of the previous for now on.

Edit: I'm going to send the notebox changes in another CSS-focused PR once I fix it up. Will mention its formatting in the CONTRIBUTING doc.

FichteFoll:
All right, I'll hold off until you give me a clear sign it's ready for re-review.

averms:
Please don't tell people how to use ordered chapters :(

stuxcrystal:
What's the status here?

OrangeChannel:
Have to clean up the code of the ordered-chapters automation script a bit before I get to this, but I should probably be able to get around to this next week. Kind of forgot about this after I became too lazy to re-write with the BDMV info, and then got distracted by a million other things :laughing: .

OrangeChannel:
So, the script (<https://github.com/OrangeChannel/ocsuite>) can now handle 

1. Trimming + splicing the audio for all episodes and repeated chapters
1. Trimming + splicing the source clip for all episodes and repeated chapters
1. Generating SUIDs for the repeated chapters
1. Writing the chapters to valid Mastroka XMLs including adjusted timestamps and references to external segments using the generated SUIDs

so I can at least start on the re-write of [Fluff's blog post](https://mod16.org/hurfdurf/?p=8) now.

However, https://github.com/mpv-player/mpv/issues/7963 is a thing that petzku brought up yesterday and I'll be waiting to see if this is a bug or not or if something should be done about it before updating the `Caveats, Assorted Traps and Various Interesting Things to Note` section since some of this stuff is outdated or doesn't apply to mpv.

EDIT: All of this stuff is limited to constant-frame-rate sources though sadly, the logic behind converting frame numbers to timestamps isn't difficult for VFR, but doing so while cutting out sections (repeated/ordered chapters) is not feasible really.

Lypheo:
>EDIT: All of this stuff is limited to constant-frame-rate sources though sadly, the logic behind converting frame numbers to timestamps isn't difficult for VFR, but doing so while cutting out sections (repeated/ordered chapters) is not feasible really.

Why not, exactly? Couldn’t you just extract the timestamps and look them up by frame number instead of calculating them?

OrangeChannel:
Actually, since the script now returns the trimmed/spliced clips, vfr chapter timestamps are possible with `vspipe`'s timecodes files, but is there a way to do this without creating a temporary script for each clip and then reading from a timecode file? I can technically do this in a really messy way with `.get_frame(n)['_DurationNum']` without breaking out of the script but this would probably be pretty complicated/ugly. @Lypheo is there a better way to get timestamps from a vfr vs clip?

Lypheo:
Not sure exactly how ocsuite operates but maybe sth like this?
```py
src = core.ffms2.Source(r"anime_tiddies.mkv", timecodes="tc.txt")
tc = [int(float(x)) for x in open("tc.txt", "r").read().splitlines()[1:]]

def fu(n,f):
	fout = f.copy()
	fout.props["TimeStamp"] = tc[n]
	return fout
	
o = core.std.ModifyFrame(src, src, fu)
```

OrangeChannel:
I've implemented a pure Python solution in acsuite for now

```py
def f2ts(f: int, /, *, precision: int = 3, src_clip: vs.VideoNode) -> str:
    """Converts frame number to a timestamp based on framerate.
    Can handle variable-frame-rate clips as well, using similar methods to that of vspipe --timecodes.
    Meant to be called as a functools.partial with 'src_clip' specified before-hand.
    """
    if precision not in [0, 3, 6, 9]:
        raise ValueError(f"f2ts: the precision {precision} must be a multiple of 3 (including 0)")
    if src_clip.fps != fractions.Fraction(0, 1):
        t = round(10 ** 9 * f * src_clip.fps ** -1)
        s = t / 10 ** 9
    else:
        s = clip_to_timecodes(src_clip)[f]

    m = s // 60
    s %= 60
    h = m // 60
    m %= 60

    if precision == 0:
        return f'{h:02.0f}:{m:02.0f}:{round(s):02}'
    elif precision == 3:
        return f'{h:02.0f}:{m:02.0f}:{s:06.3f}'
    elif precision == 6:
        return f'{h:02.0f}:{m:02.0f}:{s:09.6f}'
    elif precision == 9:
        return f'{h:02.0f}:{m:02.0f}:{s:012.9f}'


@functools.lru_cache
def clip_to_timecodes(src_clip: vs.VideoNode) -> collections.deque:
    """Cached function to return a list of timecodes for vfr clips."""
    timecodes = collections.deque([0.0], maxlen=src_clip.num_frames + 1)
    curr_time = fractions.Fraction()
    for frame in src_clip.frames():
        curr_time += fractions.Fraction(frame.props['_DurationNum'], frame.props['_DurationDen'])
        timecodes.append(float(curr_time))
    return timecodes
```
as this should be used only on source filter inputs, the speed is really limited by the current implementation of `VideoNode.frames()` which according to stux, will be *much* faster once the pending frames PR gets merged.

The speed is not really great (~50 seconds to get 100 timestamps from a 1 million long frame vfr clip created by `core.std.BlankClip`) and deque only seems to improve performance by around 2% at most, but this will hopefully be a lot faster when the new frames generator is implemented.

Lypheo:
But why though 🤔 . With ffms2's ``timecodes``, you incur no speed penalty at all, and afaict it should be identical in funtionality.

OrangeChannel:
> But why though :thinking: . With ffms2's `timecodes`, you incur no speed penalty at all, and afaict it should be identical in funtionality.

_For the audio cutting script this is actually part of:_
Hmmm, this does require the vfr clip to be directly read from a file though, so wouldn't work for clips that somehow become vfr via vapoursynth filtering, but thinking about it... this shouldn't really ever be a case since *this* is for cutting audio which will likely be synced with the source anyways.

Also, the way to implement this would require the user to supply a timecodes file that they would need to generate (assuming they have ffms2 installed and loaded), as the current functionality takes in a VideoNode as input, not a filename I could open with ffms2.Source(). I *could* make it so a provided timecodes file will take precedence over the `clip_to_timecodes` function, only using it as a fallback if only a clip is given. For the ordered chapters script, I could also just make the audio and write_to_xml methods take a timecodes filename as a argument, but I don't know if I want to force the user to re-generate an index + a timecodes file if they've used a different source filter to get the vfr clip in the first place.

_For the chapter timestamp part of the ordered chapters creation script:_
Now, on to the real reason using existing frame props here in *ac*suite is actually a decent solution for the ordered chapters script:
Once I have the vfr clips split along chapter boundaries, I stitch them back together using `std.Splice`, allowing me to just grab frame numbers and timestamps from the already adjusted clips, without having to do the math to adjust frame timestamps to begin at zero and be continuous across cut-out repeated chapters (although technically this wouldn't be that difficult, I think).

OrangeChannel:
Just did some testing with an actual mkv clip (34k frames loaded via ffms2), and it is unusably slow :disappointed: , also noticed that ffms2 timecodes round to nearest ms, which is not ideal for chapter timings but I'm not sure the precision really matters at all at that level. Assuming the frames PR even gives this a massive speed-up, it would have to be like 2 orders of magnitude faster than it is now for this to be usable on very long clips (i.e. merged clips from a BDMV). Also found out Splice complains when trying to add a BlankClip with a loaded clip with different frame rates (using the `+` operator) but not for two clips with differing framerates created by std.BlankClip :thinking: . Looks like using a timecodes file is going to be necessary for the ordered chapters script, so I'm going to start refactoring stuff now so the timestamp calculation doesn't just use clip.fps and the functools.reduce operation won't error out for differing framerate segments.

Lypheo:
> Also found out Splice complains when trying to add a BlankClip with a loaded clip with different frame rates (using the + operator) but not for two clips with differing framerates created by std.BlankClip 🤔 

The loaded clip probably differed in format or dimensions. There's a bug in ``Splice`` that I never bothered to report which causes it to state fps mismatch as the error cause even though the format/res was the problem.

OrangeChannel:
> The loaded clip probably differed in format or dimensions. There's a bug in `Splice` that I never bothered to report which causes it to state fps mismatch as the error cause even though the format/res was the problem.

Ahh, that makes a lot more sense actually, still giving Splice a list of segments is actually cleaner looking than a `functools.reduce(operator.add, segments)` anyways so I went ahead and changed it.

I still can't think of a good way to use timecodes files to calculate the new frame times once the repeated chapters are cut out besides the method I'm using now, which turns out to be extremely slow. VFR clips are now technically supported in both the audio and ordered chapters scripts, though it will be *very* slow to write the chapters to xml if the clip is vfr.

PR #55, opened 2020-02-17
Add infoxboxes

OrangeChannel:
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -358,7 +358,7 @@ separated by a horizontal rule `---`,
 where `#` is to be replaced with an increasing
 and per-page unique number.
 
-<div class="warningbox"><p>
+<div class="warning box"><p>
 The footnote text at the bottom of the page
 must all be written one line per footnote.
 Line breaks <strong>cannot</strong> be used.
@@ -381,14 +381,14 @@ These paragraphs must be separated with a blank line above and below similar
 to images or code blocks.
 
 ```md
-<div class="infobox"><p>
+<div class="info box"><p>
 Text here as usual, using semantic linefeed rules.
 If you need text-formatting, you <strong>must</strong> use in-line HTML.
 </p></div>
 ```
 
 ```md
-<div class="warningbox"><p>
+<div class="warning box"><p>
 This class should be used for important information.
 </p></div>
 ```
@@ -412,7 +412,7 @@ $$
 Similar to \`\`\` fenced code blocks,
 separate these blocks with one blank line on either side.
 
-<div class="infobox"><p>
+<div class="info box"><p>
 Both {<code>% math %</code>} and {<code>% endmath %</code>} templates
 can be entirely replaced by two <code>$</code>'s,
 so any math between \$\$ ... \$\$ will render with KaTeX<sup><a href="#fn_3" id="reffn_3">3</a></sup>.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -363,6 +363,25 @@ must all be written one line per footnote.
 Line breaks **cannot** be used.
 
 
+### Info boxes
+
+Info boxes can be used similarly to footnotes,
+but for information that is important for the reader to understand
+before continuing to read the rest of the page.
+
+The current syntax uses in-line HTML to render these paragraphs with a
+different CSS style.
+These paragraphs must be separated with a blank line above and below similar
+to images or code blocks.
+
+```md
+<div class="infobox"><p>
+Text here as usual, using semantic linefeed rules.
+If you need text-formatting, you <strong>must</strong> use in-line HTML.
+</p></div>
+```
+
+
 ### Mathematics with TeX
 
 This guide has KaTeX support,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -418,7 +418,8 @@ can be entirely replaced by two <code>$</code>'s,
 so any math between \$\$ ... \$\$ will render with KaTeX<sup><a href="#fn_3" id="reffn_3">3</a></sup>.
 However, using \$\$ within in-line `code`, code blocks, or info boxes
 will change these into the templates above
-(making printing "\$\$" in code impossible).</p></div>
+(making printing "\$\$" in code impossible).
+</p></div>
 
 [KaTeX documentation]: https://katex.org/docs/supported.html
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -421,4 +421,4 @@ will change these into the templates above
 [^3]: Please view the [markdown of this page][contrib-md] for an example of KaTeX math using `$` symbols.
 
 [new-gitbook]: https://www.gitbook.com/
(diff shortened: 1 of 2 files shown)


OrangeChannel:
@FichteFoll what's your opinion on there being two versions of this "box", one with the current set-up (I still prefer italics for notes but I'll go with what you think looks right) and one using <https://fontawesome.com/icons/exclamation-circle?style=solid>? It would be slightly different style-wise, 100% opacity and maybe a thicker border or a shadow? I'm suggesting this because there's a couple examples of paragraphs that are currently styled the same even though one is a "side-note" and one is like a "**this is important to know**" type of thing?

CONTRIBUTING.md
```diff
@@ -358,9 +358,30 @@ separated by a horizontal rule `---`,
...
+
+The current syntax uses in-line HTML to render these paragraphs with a
+different CSS style.
+These paragraphs must be separated with a blank line above and below similar
+to images or code blocks.
+
+```md
+<div class="infobox"><p>
+Text here as usual, using semantic linefeed rules.
+If you need text-formatting, you <strong>must</strong> use in-line HTML.
```
FichteFoll, line 386: I believe you can use in-line Markdown again if you have a blank line after the opening sequence of `<div … p>` – or at least that's how it works in Github's parser. Please test first.

CONTRIBUTING.md
```diff
@@ -399,4 +421,4 @@ will change these into the templates above
 [^3]: Please view the [markdown of this page][contrib-md] for an example of KaTeX math using `$` symbols.
 
 [new-gitbook]: https://www.gitbook.com/
-[contrib-md]: https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/edit/master/CONTRIBUTING.md
+[contrib-md]: https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/edit/master/CONTRIBUTING.md#mathematics-with-tex
```
FichteFoll: That anchor doesn't work for me. Besides, wouldn't you rather want to link to the raw page?

styles/website.css
```diff
@@ -1716,3 +1713,33 @@ img {
...
+    padding: .5em .5em .5em 0;
+    background: 0 0;
+    width: 100%
+}
+.infobox p {
+    padding-top: .5em;
+    font-size: .95em;
+    margin-bottom: 0;
+    padding-bottom: .5em
+}
```
FichteFoll: If the `p` tag is just used for spacing, do we need it at all? Or, alternatively, if we do need the `p`, could we get rid of the `div`?

FichteFoll (commented):
Thanks for splitting this into a separate PR. The box looks good to me and I believe we'd definitely like to have a separate icon and color coding for warning (or important boxes, even).

CONTRIBUTING.md
OrangeChannel: Oh, I had forgot the way github handles that is different from jekyll. And yea, probably will revert this since linking to the whole page works just fine.

styles/website.css
OrangeChannel: I think the reason I split this while testing was the padding was messing up the location of the info symbol, I can test merging them together when I revisit this. If it doesn't work combined, some `!important`s for the `::before` might help. I *think* getting rid of the `div` will mean the border only surrounds the text and not the symbol, but I guess moving `.infobox::before` to `.infobox p::before` would fix that if that's even possible.

OrangeChannel: Also, you might've noticed I split the border-radius command out because I was playing around with different border radius's on the left and different styles and wanted to keep it split until I was satisfied with the style. Right now they all have the same value so it's redundant, but just an fyi as to why those are split.

CONTRIBUTING.md
OrangeChannel: Doesn't seem to work. Still displays `*text*` as `*text*` and not *text*.

styles/website.css
OrangeChannel: Merged in the `p` tag in https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/pull/55/commits/76975702f2301f3c588cb4f66955a8b4b0f69b9c and got rid of all the redundant code.

OrangeChannel:
Here's what the new formatting looks like:

image
image
image

styles/website.css
```diff
@@ -1713,28 +1713,32 @@ img {
         display: none;
     }
 }
-.infobox {
+.infobox,.warningbox {
```
FichteFoll: I believe in this case it's better to use `.box.info`, i.e. two classes, to clearly indicate their shared layout.

CONTRIBUTING.md
```diff
@@ -383,6 +387,12 @@ If you need text-formatting, you <strong>must</strong> use in-line HTML.
 </p></div>
 ```
 
+```md
+<div class="warningbox"><p>
+This class should be used for important information.
+</p></div>
```
FichteFoll: Didn't the last commit remove the need for the p tag?

FichteFoll:
Visuals are good, especially how you made them visually distinct without using a different color. I hope that's enough to distinguish them for other eyes as well.
However, I don't really like the increased border radius for the warning boxes personally.

CONTRIBUTING.md
FichteFoll: Sad!

CONTRIBUTING.md
OrangeChannel: It removed the need for specifying a new style for this div's paragraph tag. You still need to use the `p` tag for it to inherit the correct styles from the rest of the page. This is what it looks like without the `<p>text</p>`. The margins/padding are missing and the text isn't displayed corectly because I guess it doesn't know that it needs normal html paragraph text-formatting (the spacing around the bold text).
image

OrangeChannel:
I had increased the border-radius because with the thicker border, it looks significantly more rectangular (?) or sharp with the `.2em` compared with the thin dashed border. I can switch both to the same (maybe both `.25em`) or lower the warning one a bit but I'll leave the decision up to you.
`.5em`:
image
`.4em`:
image
`.2em` (same as infoboxes):
image

styles/website.css
OrangeChannel: This will mean that in-line it will have to be specified as `<div class="info box"><p>` and `<div class="warning box"><p>` with the space between the words, but I agree it makes it clearer.

Example css:
```css
.box {
    display: inline-flex;
    padding: .25em .75em .25em 0;
    border-left: .4em solid!important;
    width: 100%
}
.box::before {
    content: "";
    font: 14px/1 FontAwesome;
    font-size: 1.75em;
    margin: 0 .5em 0 .75em;
    align-self: center
}
.box.info {
    border: .1em dashed;
    border-radius: .2em;
    opacity: 80%;
    font-size: .95em
}
.box.warning {
    border: .2em outset;
    border-radius: .5em; /* this can be changed */
    box-shadow: 0 0 .2em rgba(0,0,0,.8)
}
.box.warning::before {
    content: ""
}
```

CONTRIBUTING.md
FichteFoll: All right, thanks for clarifying.

PR #56, opened 2020-03-02
Rewrite descaling guide from scratch

LightArrowsEXE:
I hated my old write-up, so here's an updated one!


OrangeChannel:
Okay, so this was a bit weird to do since it's coming from @LightArrowsEXE 's master branch, but I've rebased this onto the current upstream master, as well as got rid of the `back-up` folder I assume created by pingo or some compression app. Some images weren't fully compressed so I've also went back and compressed those.

- Original (with back-up folder): `71225520 B`
- Original (without folder): `19175192 B`
- Current (no folder, new compression): `14956763 B` 79% saved total, 22% saved with new compression

---
EDIT: not sure how to give Light back the ownership of that commit (if that's even possible)
Here's a backup of the original commit <https://github.com/OrangeChannel/guide.encode.moe/tree/descale-rewrite-backup>

FichteFoll:
FWIW, you can set the author of a commit with `git commit --author="FichteFoll <fichtefoll2@googlemail.com>"`. I'll do that before merging.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+When performing a simple downscale,
+information will be lost in the process.
+A proper descale will, on the other hand,
+(theoretically) preserve all the *actual* detail in an image.
+
+That said, if you can't perform a good descale,
+you will usually end up with haloing/ringing artifacting,
+as illustrated below:
+
+![](images/descale/descale_ubw_00_bad.png)
```
FichteFoll: Please add a caption to every image. I learded to appreciate this style in my academic work and would like to maintain it for this project as well.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+used for figuring out the native resolution of an image.
+It will return a graph showcasing all the resolutions
+it tried to descale to,
+as well as a txt file
+with all the relative errors listed.
+Here's an example:
+
+![](images/descale/descale_owari_01_frame.png)
+
+![getnative.py -k bicubic -b 0 -c 1/2](images/descale/descale_owari_01_graph.png)
```
FichteFoll: Please also add descriptive captions for these two. For the latter, do something like `The result of …`. I'm not going to repeat this for the cases below.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+
+What you want to look out for are very clear spikes.
+You can spot one of those here,
+pointing towards a height of 720.
+When trying to run getnative,
+you ideally want to get a relatively bright frame
+with very clearly defined lineart
+and as little effects or filtering applied to it,
+like for example blur or glow.
```
FichteFoll: I wonder if it's enough to just say this or whether an example image would help.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+In what way does it differ from normal downscaling?
+How do you actually descale properly?
+This guide will explain the groundwork for you
+to start descaling effectively.
+
+
+## Descaling vs. Downscaling
+
+What the difference between downscaling and descaling is
+is a question that gets asked fairly frequently.
```
OrangeChannel, line 15: I don't think this fits here. Just explaining it directly works.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+Simply put,
+downscaling is the act of resizing an image
+to a smaller resolution than the original image,
```
OrangeChannel, line 18: I think we should have some mention of how downscaling uses the same kernels as upscaling, while de-scaling reverses the kernel's logic. We can reference the resampling page here as well.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+while with descaling you are actively reversing the math
+applied to an upscale,
+reconstructing the original image before upscaling
+(which coincidentally also means you get a smaller image).
```
OrangeChannel: Probably unnecessary to write this.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+
+For a more thorough explanation on the intrinsics of descaling,
+please refer to the [vapoursynth-descale Github page][vapoursynth-descale].
+
+Descaling is generally the preferred method
+of handling upscaled content whenever possible.
+When performing a simple downscale,
+information will be lost in the process.
+A proper descale will, on the other hand,
+(theoretically) preserve all the *actual* detail in an image.
```
OrangeChannel: Explain here that `*actual*` means the detail that was in the image before it was upscaled. (Also maybe talk about how due to MPEG compression and blocking done *after* the upscale, some information is distorted or non-recoverable.)

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+you will usually end up with haloing/ringing artifacting,
+as illustrated below:
+
+![](images/descale/descale_ubw_00_bad.png)
+
+<div class="warning box"><p>
+Only descale if you can be sure,
+beyond reasonable doubt,
+that you have figured out the correct resolution and kernel.
+</p></div>
```
OrangeChannel, line 51: I think we need to first explain how to get the correct resolution and kernel before including this warningbox, or at least start with a header that mentions this (fits better at the top of the next section IMO).

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+
+[vapoursynth-descale]: https://github.com/Irrational-Encoding-Wizardry/vapoursynth-descale
+
+
+# Discerning the native resolution
+
+The first step to descaling is to figure out
+what the native resolution of a frame is.
+One way to do this is by descaling to a bunch of resolutions,
```
OrangeChannel:
```suggestion
One way to do this is by descaling to all potential resolutions,
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+the artifacts created by using the wrong kernel.
+Luckily for us,
+there's already a tool out there
+you can use for this.
+
+[getnative.py][getnative] is a common tool
+used for figuring out the native resolution of an image.
+It will return a graph showcasing all the resolutions
+it tried to descale to,
+as well as a txt file
```
OrangeChannel:
```suggestion
as well as a plaintext file
```

encoding/descaling.md
OrangeChannel:
```suggestion
When using getnative,
the ideal frame to use is a relatively bright one
with clearly defined lineart
and no obvious effects or filtering applied to it
(blur or glow).
```
Doesn't grain also appy here as an "unwanted effect"?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+
+![](images/descale/descaled_fgo_op_encoded_frame.png)
+
+![Running getnative on frame of a already-rescaled encode](images/descale/descaled_fgo_op_encoded_graph.png)
+
+Also of note is that you should be careful
+to not pick a frame with credits to run getscaler on.
+Credits are usually added at the very end,
+after the clip has already been upscaled.
+Those will need to be handled separately.
```
OrangeChannel:
```suggestion
Don't use getnative on a frame with credits or text.
Credits are usually added at the very end of the Blu-ray creation process,
after the clip has already been upscaled.
These will need to be handled separately.
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+
+![](images/descale/descaled_yaiba_17_credits_frame.png)
+
+![Credits and other native 1080p elements mess with the graph](images/descale/descaled_yaiba_17_credits_graph.png)
+
+And last but not least,
+we have letterboxing.
+Letterboxing is what we call the black bars surrounding the video.
```
OrangeChannel:
```suggestion
Letterboxing in video used to be the result of transferring a film
shot in a widescreen aspect ratio to a standard modern video format,
while keeping the original aspect ratio.
This resulted in "black bars" above and beneath the frame.
Today, this is mostly a stylistic choice as letterboxing and windowboxing
(having black bars on all four sides)
is no longer necessary with digitally produced content.
```
Maybe we can put some of this in an infobox? Or a footnote?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+As an added bonus,
+this example is also from a blu-ray release
```
OrangeChannel:
```suggestion
this example is also from a Blu-ray release
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-![Manaria Friends — 01 (frame 3)](images/descale_manaria03.png)
+clip_y = core.std.ShufflePlanes(clip, 0, vs.GRAY)
+descaled = core.descale.Debicubic(clip_y, width=1280, height=720, b=1/3, c=1/3)
+reupscaled = core.resize.Bicubic(descaled, width=1920, height=1080, filter_param_a=1/3, filter_param_b=1/3)
+merged = core.std.ShufflePlanes([reupscaled, clip], [0, 1, 2], vs.YUV)
 
-Heavy dynamic grain will almost always give bad results.
+comp = lvf.compare(clip, merged)
+comp.set_output()
+```
```
OrangeChannel:
```py
import descale

src = core.lsmas.LWLibavSource(r'/path/to/file.m2ts')

descaled = descale.Debicubic(src, width=1280, height=720, b=1/3, c=1/3)
reupscaled = core.resize.Bicubic(descaled, width=1920, height=1080, filter_param_a=1/3, filter_param_b=1/3)

core.std.Interleave(clips=[src.text.Text('src'), reupscaled.text.Text('scaled', 9)]).set_output()
```
Is there a reason here for changing the bitdepth? And using the python wrapper for descale you don't need to convert the clip or merge the planes back.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-but it isn't full of effects
-and is fairly bright.
-The lineart is very clear.
+![](images/descale/descale_zoom_a.png)
 
-We will now make use of the getnative.py script
-to figure out what resolution
-this anime was produced at.
-Run the following in your terminal:
+Clip "No zoom" and select "Fixed ratio".
```
OrangeChannel:
```suggestion
Click "No zoom" and select "Fixed ratio".
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-This is a good result
-for testing out varying kernels.
 
+A couple more extreme examples
+would be using [waifu2x][] or [upscaled_sraa][]
+to upscale it back to 1080p.
+These are far more destructive
+and take much longer,
+but depending on your source,
+might yield overall better results.
```
OrangeChannel:
```suggestion
These are far more destructive and expensive to compute
but might yield overall better results
depending on your source.
```

OrangeChannel (changes requested):
I've only gone through this up until the ` Handling native 1080p content` since I know too little of how the credit stuff is done.

encoding/descaling.md
FichteFoll: > how downscaling uses the same kernels as upscaling

That's … a pretty general thing to say and doesn't exactly provide more information than what we have currently.

encoding/descaling.md
FichteFoll: Not entirely unnecessary, but the "coincidentally" is too strong. It was our intention all along to go back to the original information and reduce the amount of superficial information that needs to be included in an encode, or provide better such superficial information.

Actually, I'm missing this kind of explanation for the motivation of descaling. You say that it's better than downscaling, but never mention why we want to do it in the first place, particularly why it's **always** a good idea to do so (before applying your filter chain), provided it's possible.

encoding/descaling.md
FichteFoll: Generally mentioning any processing done after the scaling, including compression artifacts, should be good. Blur is mentioned later, for example, but this advice can be general.

encoding/descaling.md
FichteFoll: The warning serves the purpose of informing the reader to only apply what they have learned so far to certain cases and with certain parameters, i.e. when they have determined the correct kernel and resolution to properly revert the upscale. That information in itself is useful and important enough without knowing how to determine these.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+beyond reasonable doubt,
+that you have figured out the correct resolution and kernel.
+</p></div>
+
+[vapoursynth-descale]: https://github.com/Irrational-Encoding-Wizardry/vapoursynth-descale
+
+
+# Discerning the native resolution
+
+The first step to descaling is to figure out
```
FichteFoll, line 58: I think we can put more emphasis on the importance of this, in accordance with the warning box earlier.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+are generally more obvious than
+the artifacts created by using the wrong kernel.
+Luckily for us,
+there's already a tool out there
+you can use for this.
+
+[getnative.py][getnative] is a common tool
+used for figuring out the native resolution of an image.
+It will return a graph showcasing all the resolutions
+it tried to descale to,
```
FichteFoll, line 73: Needs a mention that it does this for a given kernel (and parameters), imo.

encoding/descaling.md
FichteFoll: Why should this be worthy of a warning box? It's about as important as choosing a frame without post-processing or letterboxing.

encoding/descaling.md
FichteFoll: Letterboxing is still necessary for non-standard aspect ratios not covered by disc formats, mostly Blu-ray nowadays. Although these hardly get produced for obvious reasons.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-too many post-processed elements in the picture.
-Whilst it is most definitely possible
-to get pretty good results with "bad" frames,
-it's generally better to use
-good frames whenever possible.
+Once you've figured out the native resolution,
+you can move on to the actual descaling.
+There's two common ways to do this.
+The first one is by using [descale][].
+The other is by using `invks` from [fmtc][].
```
FichteFoll: Could use an ordered list here.

```suggestion
There's two common ways to do this:

1. Using the [descale][] plugin.
2. Using `invks` from the [fmtc][] plugin.

```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+For convenience's sake,
+we'll be keeping it to just `descale` for this guide.
 
-Here are some examples of "bad" frames.
+There are a couple very common kernels
+used in anime production:
+* Bilinear
```
FichteFoll: Please add blank lines around lists, ordered or unordered.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+Occasionally, although rare,
+you may also run into the following kernels:
+* Lanczos taps=3
+* Lanczos taps=4
+* Spline16
+* Spline36
+* Bicubic b=1, c=0 (B-Spline)
+* Bicubic b=0, c=0 (Hermite)
+* Bicubic b=0, c=0 (Hermite)
+* Bicubic b=0.3782, c=0.3109 (Robidoux)
```
FichteFoll, line 170: I've never heard of Robidoux, but I'm not an encoder. Are these actually used?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-![Manaria Friends — 01 (frame 1)](images/descale_manaria01.png)
+Occasionally, although rare,
+you may also run into the following kernels:
+* Lanczos taps=3
+* Lanczos taps=4
+* Spline16
+* Spline36
+* Bicubic b=1, c=0 (B-Spline)
+* Bicubic b=0, c=0 (Hermite)
+* Bicubic b=0, c=0 (Hermite)
```
FichteFoll:
```suggestion
* Bicubic b=0, c=0 (Hermite)
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+* Bicubic b=0.2620, c=0.3690 (Robidoux Sharp)
+* Bicubic b=0.6796, c=0.1602 (Robidoux Soft)
 
-This picture is dark.
-It also has some effects over it.
+The most reliable way to discern
+which kernel was used is by descaling the clip,
+upscaling it again with the same kernel,
+and then comparing it with the original frame.
+Here is an example script using `compare` from `lvsfunc`
```
FichteFoll: Link for `lvsfunc`?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-Heavy dynamic grain will almost always give bad results.
+comp = lvf.compare(clip, merged)
+comp.set_output()
+```
 
-![Manaria Friends — 01 (frame 4)](images/descale_manaria04.png)
+The most reliable way to check for differences is
+by zooming in 4x with Nearest Neighbor
+and switching between the frames
+with the left and right arrow buttons.
```
FichteFoll, line 195: Is this using vsedit? Should probably mention that, since yuuno has different means of comparing clips.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-![Miru Tights — 02 (getnative graph)](images/descale_bad_graph1.png)
+# Re-scaling
 
-```
-Output:
-Kernel: bicubic AR: 1.78 B: 0.33 C: 0.33
-Native resolution(s) (best guess): 869p, 848p
-```
+A fairly common practice is
+to upscale the descaled clip to a standard resolution whenever possible
```
FichteFoll: "whenever possible" is ambiguous. When is it possible? What could prevent one from resizing to a standard resolution?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+(if it isn't already one, like for example 720p).
+There's a couple scalers that are generally recommended.
```
FichteFoll:
```suggestion
The following scalers are generally recommended for this task.
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
 
-Another pitfall you've got
-to watch out for
-is checking the results of a frame
-with letterboxing.
+Similarly, [nnedi3_resample][] is also used by some people
+and might give better results
+with some additional tweaking of the settings.
+Like nnedi3_rpow2, it's an image doubler
+and needs to be downscaled after.
```
FichteFoll, line 244:
```suggestion
and needs to be downscaled afterwards.
```

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-there are no clear spikes in this graph
-like there was in the Manaria Friends one.
-The results are not clear enough
-to work off of.
-Here's another example:
+The most common scaler for re-scaling is [nnedi3_rpow2][].
+It returns consistently good-looking lines
+without damaging other detail.
+Since it's an image doubler,
+it's important to remember that you'll have to scale it down afterwards.
```
FichteFoll, line 230: No recommendation for a downscaling kernel?

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+is by using the re-upscaled clip
+used for checking for the kernel
+and using an expression to get the absolute difference between the two
+and then tweaking the mask further to properly catch everything.
+Then the mask is binarized
+so it doesn't catch every small extra bit
+it happened to catch.
+This should work fine for most of the frame
+(given that you descaled it correctly).
+Here is an example of that:
```
FichteFoll, line 304: Not happy with this paragraph. It's a bit too unprecise imo and worded poorly. On one hand, you describe the steps required, but on the other you leave out important parts of the steps. Like, you say to "tweak" the mask (which at this point is just the difference between two clips) instead of expanding it.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
+descaled = core.descale.Debicubic(clip_y, 1550, 872, b=1/3, c=1/3)
+upscaled = core.resize.Spline36(descaled, 1920, 1080, filter_param_a=1/3, filter_param_b=1/3)
+credit_mask = core.std.Expr([clip_y, upscaled], 'x y - abs')
+credit_mask = kgf.iterate(credit_mask, core.std.Maximum, 4)
+credit_mask = kgf.iterate(credit_mask, core.std.Inflate, 2)
+credit_mask = credit_mask.std.Binarize(0.05)
 
-There is one very, very important thing
-to keep in mind when descaling:
+rescaled = nnedi3_resample(descaled, 1920, 1080)
```
FichteFoll: `nnedi3_resample` is undefined in this script.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-For example,
-here it was used
-specifically to fix up
-some bad aliasing
-in the source.
-
-![Akanesasu Shoujo — 01 (src)](images/descale_akanesasu_src.png)
-
-![Akanesasu Shoujo — 01 (rescaled)](images/descale_akanesasu_rescaled.png)
+![Merge the original credits with a re-upscaled frame (using nnedi3_resample)](images/descale/descaled_symphoxv_credits_frame_masked.png)
```
FichteFoll: I don't see the image here, but is it a composition or the final frame? In the latter case, the caption can't use imperative.

encoding/descaling.md
```diff
@@ -1,477 +1,332 @@
...
-on the CGI model.
-
-***
+![](images/descale/descale_casefiles_ed_credits_frame_non-masked.png)
 
-[^1]: Most, if not all, relevant VapourSynth scripts/plug-ins and their functions can be found in the [VapourSynth Database][vsdb].
+![Downscaled mask that catches all the credits](images/descale/descale_casefiles_ed_credits_mask.png)
 
-[vsdb]: http://vsdb.top/
+![Merge downscaled credits with descaled clip](images/descale/descale_casefiles_ed_credits_frame_masked.png)
```
FichteFoll, line 360: Same as above.

Also, maybe a slowpics link for these images would be useful?

FichteFoll (changes requested):
I like the flow in the rewrite a lot better than in the previous. Good job on that.

Please address the comments I added and also what I didn't comment on in Dave's review.

encoding/descaling.md
LightArrowsEXE: The reason for why you should descale gets mentioned in the next paragraph. I'll build onto that a bit further.

encoding/descaling.md
LightArrowsEXE: I agree with Fichte here. It's already established before what kind of frames you should pick.

encoding/descaling.md
LightArrowsEXE: The reason for letterboxing being a thing is also irrelevant for descaling. We're talking about preparing a frame to be descaled, not the history of letterboxing.

encoding/descaling.md
LightArrowsEXE: Yup, my bad. These are examples of the plugin, not the wrapper. Will update.

encoding/descaling.md
LightArrowsEXE: I personally find that rather annoying to do when a lot of it is mostly repeats of the same thing and is already pretty evident just by reading the accompanying text. Maybe the first time around, but for others? Ehhh...

encoding/descaling.md
LightArrowsEXE: I think it is, but I'll add an example image and graph just in case.

encoding/descaling.md
LightArrowsEXE: In the example code I use the internal scaler (which is Spline36), and it's also what I'd recommend either way. I'll note that Nnedi3_rpow2 (and resample) can downscale a clip inherently.

encoding/descaling.md
LightArrowsEXE: I don't quite understand this comment. What are we supposed to put more emphasis on? That you need the correct resolution? Because I can't think of anything that puts more emphasis on it than writing an entire section on it.

encoding/descaling.md
LightArrowsEXE: With the default (bicubic 1/3 1/3) usually returning the correct resolution either way, I don't think it's necessarily worth pointing that out.

encoding/descaling.md
LightArrowsEXE: They've ended up being the best results for some stuff I've come across.

encoding/descaling.md
LightArrowsEXE: I don't quite get what you mean by this. Could you elaborate?

encoding/descaling.md
OrangeChannel: Hm... :thinking: I guess you're right, a more detailed explanation wouldn't belong here I guess but maybe a `## Descaling` section on the resampling page may be something to consider in the future (that better explains the difference mathematically).

LightArrowsEXE:
Have updated it with the comments in mind. I may have missed one or two, so another double-check would be appreciated. I've also replied to some comments that I either wanted elaboration on or don't agree with.

encoding/descaling.md
OrangeChannel: Updated suggestion with above feedback.

LightArrowsEXE: They're not added at the end of the "Blu-ray creation process". They're added near the end of the _mastering_. I'll update it with that and push that alongside additional suggestions made on the new commit.

OrangeChannel:
@FichteFoll I caught light up with the changes so any further changes shouldn't be a problem (sorry for the double push, git deleted+added files instead of renaming them and i didn't notice).

Before merging, just make Light author of 03732aba37a092c62362b0ac8ef962b313383168﻿ and 9f0eed5a31f8df25b308afe7c78b07673954928d .

encoding/descaling.md
OrangeChannel: Technically resolved this issue but there's another undefined thing, I'll make a separate suggestion.

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
+A proper descale will, on the other hand,
+(theoretically) preserve all the detail
+that was actually there in the original image.
+This offers a multitude of advantages,
+like for example when dealing with artifacting
+that was a result of a poor upscale.
+You could instead undo the scaling
+and use a better scaler to upscale it instead,
+or keep it at its native resolution and
+have less artifacting to worry about.
```
OrangeChannel, line 36:
```suggestion
This offers many advantages,
one example being the ability to fix artifacting
that was a result of a poor upscale.
You could undo the scaling
and use a better scaler to upscale it instead,
or keep it at its native resolution and
have less artifacting to worry about.
```
Are there any other benefits worth mentioning?

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
+[vapoursynth-descale]: https://github.com/Irrational-Encoding-Wizardry/vapoursynth-descale
+
+
+# Discerning the native resolution
+
+The first step to descaling is to figure out
+what the native resolution of a frame is.
+One way to do this is by descaling to all potential resolutions,
+upscaling the descaled image with the same kernel,
+and comparing the two.
```
OrangeChannel, line 62:
```suggestion
and comparing this result with the original.
```

encoding/descaling.md
OrangeChannel: Maybe adding a footnote mentioning:

> With the default (bicubic 1/3 1/3) usually returning the correct resolution either way...

would satisfy this?

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
+it tried to descale to,
+as well as a plaintext file
+with all the relative errors listed.
+Here's an example:
+
+![](images/descale/descale_owari_01_frame.png)
+
+![Results from "getnative.py -k bicubic -b 0 -c 1/2"](images/descale/descale_owari_01_graph.png)
+
+What you want to look out for are very clear spikes.
```
OrangeChannel, line 82:
```suggestion
What you want to look out for are very clear downward spikes.
```
I've seen a few awful graphs in the #bot channel that have spikes somehow going up, and we'd want to avoid these afaict.

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-![Manaria Friends — 01 (frame 1)](images/descale_manaria01.png)
 
-This picture is dark.
-It also has some effects over it.
+# Filterchains
 
-![Manaria Friends — 01 (frame 2)](images/descale_manaria02.png)
+Once you've figured out the native resolution,
+you can move on to the actual descaling.
+There's two common ways to do this.
```
OrangeChannel, line 144:
```suggestion
There's three common ways to do this.
```
Whoops, didn't notice that when I added mention of the python functions.

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-
-![Miru Tights — 02 (getnative graph)](images/descale_bad_graph1.png)
+descaled = lvf.test_descale(clip, height=720, kernel='bicubic', b=1/3, c=1/3)
 
-```
-Output:
-Kernel: bicubic AR: 1.78 B: 0.33 C: 0.33
-Native resolution(s) (best guess): 869p, 848p
+comp = lvf.compare(clip, descaled)
+comp.set_output()
```
OrangeChannel, line 189:
```suggestion
lvf.compare(clip, descaled).set_output()
```
I think chaining in this case is actually more elegant. Just an IMO suggestion, not a fix.

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
 
-![Kizumonogatari I (getnative graph)](images/descale_ararararagi_graph.png)
+# Re-scaling
 
-You will have to crop them beforehand
-or they will return odd graphs like this.
+A fairly common practice is
+to upscale the descaled clip to a standard resolution
+(if it isn't already one, like for example 720p).
+The following scalers are generally recommended for this task.
```
OrangeChannel, line 224: These seems like there's going to be a list after this line, so maybe needs some rewording.

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-Native resolution(s) (best guess): 810p
-```
-The results are very clear.
-There are a couple of deviations,
-but there's a very clear spike
-going down to 810p.
-This is a good result
-for testing out varying kernels.
 
+Similarly, [nnedi3_resample][] is also used by some people
```
OrangeChannel, line 240:
```suggestion
Another scaler, [nnedi3_resample][], is sometimes used
```

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-using other kernels.
+```py
+upscaled = core.resize.Lanczos(descaled, width=1920, height=1080, filter_param_a=3)
+```
 
-Here is a list
-of some common kernels and values.
+A couple more extreme examples
+would be using [waifu2x][] or [upscaled_sraa][]
+to upscale it back to 1080p.
```
OrangeChannel, line 265:
```suggestion
to upscale it back to the original resolution.
```

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-[manaria_compare]: https://slowpics.org/comparison/61e39e1e-d074-4d83-b7c9-b0f4e1861855
 
+# Handling native 1080p content
 
-## Credits and other native 1080p elements
+When descaling,
+it's important to be wary of native 1080p elements.
+A common example of this is credits,
+which are usually added after upscaling.
+These will require being masked.
```
OrangeChannel, line 292:
```suggestion
These will need to be masked.
```

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-
-![Akanesasu Shoujo — 01 (src)](images/descale_akanesasu_src.png)
-
-![Akanesasu Shoujo — 01 (rescaled)](images/descale_akanesasu_rescaled.png)
+[Here][rescaled] is a link to a comparison.
+
+This can also be done when descaling
+to a resolution that you plan to stick with.
+Since the credits will still be messed up,
+you will need to replace them with a downscaled clip.
```
OrangeChannel, line 336: The differentiation here is that you are going to *downscale* the credits while *descaling* the rest of the frame, right? I think better wording would be as follows:
```suggestion
Since there is no need to *descale* native 1080p elements,
you will need to *downscale* the credits instead.
```

OrangeChannel (changes requested):

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-For example,
-here it was used
-specifically to fix up
-some bad aliasing
-in the source.
-
-![Akanesasu Shoujo — 01 (src)](images/descale_akanesasu_src.png)
-
-![Akanesasu Shoujo — 01 (rescaled)](images/descale_akanesasu_rescaled.png)
+[Here][rescaled] is a link to a comparison.
```
OrangeChannel, line 331: I think that Fichte has said before not to have a link on words like "here" and instead just link the actual thing. i.e.:
```suggestion
[Slowpics comparison of the re-upscaled frame.][rescaled]
```

encoding/descaling.md
```diff
@@ -1,477 +1,365 @@
...
-on the CGI model.
+![](images/descale/descale_casefiles_ed_credits_frame_non-masked.png)
+
+![Downscaled mask that catches all the credits](images/descale/descale_casefiles_ed_credits_mask.png)
 
-***
+![Merge downscaled credits with descaled clip](images/descale/descale_casefiles_ed_credits_frame_masked.png)
 
-[^1]: Most, if not all, relevant VapourSynth scripts/plug-ins and their functions can be found in the [VapourSynth Database][vsdb].
+[Here][downscaled] is also a link to a comparison.
```
OrangeChannel, line 362: Same as previous one:
```suggestion
[Slowpics comparison of the descaled frame.][downscaled]
```

encoding/descaling.md
OrangeChannel: I think since the graphs apply to the images usually directly above, the way it renders only the graph's caption and not one for the frame actually looks better on the website.

encoding/descaling.md
OrangeChannel: I think Fichte means the caption should sound more like `Original credits merged with a re-upscaled (nnedi3_resample) frame` instead of using a command-like phrase.

encoding/descaling.md
FichteFoll: I think it's just the "first step" wording that I don't like here, because it's not just the first but the most important one.

encoding/descaling.md
FichteFoll: SGTM

encoding/descaling.md
FichteFoll: Should those even be separate entries?

FichteFoll:
It's nice that github added their review feature, but things quickly become a mess for larger PRs like this. It's not easy to see all unresolved discussions and compare the changes since then to determine whether they have been resolved in code.

LightArrowsEXE:
This PR is going to be closed now and reworked later. I have a plan for a different page to be added, and will be writing a guide for that instead now. Will revisit updating descaling at a later date.

PR #57, opened 2020-03-08
video-artifacts.md: Replace Levels with resize and use a more instructive example

Lypheo:
As pointed out by Aicha, full range range adjustments are nonsensical. Also, fixing the simplest and most common of range issues, that is, incorrectly tagged full range, doesn’t require std.Levels, so add a seperate example for that.


OrangeChannel:
Just did a small formatting change of where the `[resize]: ...` link was and made commit message more specific since this won't be under a merge commit.

FichteFoll (approved):
Thanks, both of you.

PR #59, opened 2020-07-06
New x264 build released as of four days ago

Velvetine:
New x264 build released as of four days ago. Also, builds are no longer separated between 8-bit and 10-bit.


FichteFoll:
Thanks.

It seems that gitbook broke out of nothing. I'll probably have to revisit this and switch to using github actions instead of travis in the process.

PR #69, opened 2021-09-18
Initial documentation on encoding with the vpX/aom family of encoders.

ILOVEPIE:
Specifically the background information and an example of how to encode with vp9.
AV1 example will be added via a later PR.


src/encoding/codecs/vpXfamily.md
```diff
@@ -0,0 +1,292 @@
...
+	- No:
+		- Is your source Full HD Quality:
+			- Yes:
+				- Is time important?
+					- Yes:
+						- VP9
+					- No:
+						- AV1
+			- No:
+				- VP9
```
shssoichiro, line 292: Why? AV1 can still achieve better compression, so I would argue a source being low quality should not be a reason to use VP9 over AV1.

ILOVEPIE:
AV1 Is inefficient for files that have a lower resolution than 1080p due to
the way that the tiles are handled. That being said this is just a
recommendation and you can encode lower res with AV1.

src/encoding/codecs/vpXfamily.md
ILOVEPIE: AV1 Is inefficient for files that have a lower resolution than 1080p due to
the way that the tiles are handled. That being said this is just a
recommendation and you can encode lower res with AV1.

shssoichiro: Fair, although there's no need to use tiles. You can get efficient multithreading using a tool like av1an, and set tiles to 1.

ILOVEPIE: Split encoding (like how av1an works) actually results in higher bitrates compared with tile multithreading.

shssoichiro: Do you have any research to support that? Logically, there's no reason that split encoding should result in higher bitrate than using tiles. It should be the opposite. A split encoded stream should be the equivalent of encoding that stream as one sequence.* Obviously that assumes that keyframe selections are identical, if they're not then that is the fault of the implementation.

*Assuming QP based encoding. Don't do split encoding with VBR.

ILOVEPIE: Think about it this way, each segment must start with a keyframe, If a keyframe isn't necessary in that location it's going to increase bitrate. Even if it is, among other things, libaom-av1 optimizes keyframe placement to occur on frames with large changes to the image. This means that the keyframe will likely be placed sub-optimally if you split-encode it, increasing the bitrate or lowering the quality. libaom-av1 has some very complex keyframe optimization techniques it uses which can be thrown off by split encoding like that. libvpx-vp9 does some similar but less-complex optimizations, so it won't be affected as much but it still can be.

ILOVEPIE: But anyways... Is there anything else I need to fix?

ILOVEPIE:
Any way this can get pulled?

OrangeChannel:
I don't have anything set up to preview this book atm, but I will try to check over the grammar and formatting this weekend.

OrangeChannel:
Looks good to me but I'm not too familiar with vpX/aom encoders so it'd be nice if we can get someone else to review the actual content.

ILOVEPIE:
Any updates on this?

FichteFoll:
No, not yet. When I find some time over a weekend, I intend to look over it again and probably merge it in its current state with some adjustments, if necessary. It's still on my radar but other things have a higher priority currently.

PR #79, opened 2023-03-01
x265: update mods and vspipe call

noizuy:
vspipe --y4m is deprecated (https://github.com/vapoursynth/vapoursynth/blob/master/src/vspipe/vspipe.cpp#L702)
the previously linked mods have blocking issues: https://slow.pics/c/2ZU6BUOW added links to mods that should be safe


FichteFoll (approved):
Thanks!

PR #85, opened 2024-03-31
Fix broken links.

tenturestept:
This required updating links to the new Vapoursynth documentation and
making a link to the DVD specification point to the Wayback Machine.

The link to http://maven.whatbox.ca:11665/resample_kernels/kernels.html
in resampling.md is dead too but wasn't replaced.

A sentence linking a list of plugins in the Vapoursynth docs was removed
since the list of plugins doesn't exist anymore.


FichteFoll (approved):
Thanks for the fixes!

